PR #6, opened 2018-05-16
#157505141 create base model classes

Dheavyman:
#### What does this PR do?
- Have base models with common attributes
- Create parent base and auditable classes with attributes to be inherited by other child classes
- N/A
- The PR provides consistent parent models for other model classes to be created
- #157505141
- N/A
#### Questions:
- N/A

### DIFF
--- a/api/models/base/auditable_model.py
+++ b/api/models/base/auditable_model.py
@@ -1,0 +1,15 @@
+from datetime import datetime
+from ..database import db
+
+
+class AuditableModel(db.Model):
+    """ Auditable model """
+
+    __abstract__ = True
+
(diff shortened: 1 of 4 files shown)


api/models/base/auditable_model.py
```diff
@@ -0,0 +1,15 @@
+from datetime import datetime
+from ..database import db
+
+
+class AuditableModel(db.Model):
```
kodero: Hey @Dheavyman, this is meant to be called `AuditableBaseModel`. Its also meant to extend `BaseModel`

Dheavyman: Ok, comments taken note of. Implementing the feedback

PR #8, opened 2018-05-17
#157505150 Model Operations

purpose50:
#### What does this PR do?
Centralise CRUD operations

* Centralise CRUD operations in model class
* Setup pytest
* Test centralised CRUD operations with pytest

* clone the branch
* create a new virtual env
* Set ```FLASK_ENV=testing```

* modified the exisiting Usermodel id parameter from ```uuid4()``` to ```uuid4```
* fix the configuration for test to reference the proper environment. The test configuration environment was set to ``` ProductionConfig ``` instead of ``` TestingConfig```

#157505150 Model Operations

screen shot 2018-05-17 at 11 41 57 am

#### Questions:
N/A


api/models/model_operations.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        Save a model instance
+
+        :return: Model instance
+        """
+
+        db.session.add(self)
+        db.session.commit()
+        return self
+    def update(self, **kwargs):
```
kodero: Shouldn't there be some space between line 14 and fifteen?

api/models/model_operations.py
```diff
@@ -0,0 +1,53 @@
+from .database import db
+from uuid import uuid4
+
+class ModelOperations(object):
+    def save(self):
+        """
+        Save a model instance
+
```
kodero: This space is un-necessary.

api/models/model_operations.py
```diff
@@ -0,0 +1,53 @@
...
+    def update(self, **kwargs):
+        """
+        update entries
+        """
+        for field, value in kwargs.items():
+            setattr(self, field, value)
+            db.session.commit()
+
+    @classmethod
+    def get_one(cls, id):
```
kodero: Rename this to `get`.

api/models/model_operations.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def delete(self):
+        """
+        Delete a model instance.
+        """
+        db.session.delete(self)
+        db.session.commit()
+
+    @classmethod
+    def get_all(cls):
```
kodero: rename this to `query` since we'll implement filtering on top of this.

api/models/model_operations.py
```diff
@@ -0,0 +1,53 @@
...
+        db.session.commit()
+        return all_entries
+ 
+    @classmethod
+    def count(cls):
+        """
+        return total entries in the database
+        """
+        counts = cls.query.count()
+        db.session.commit()
```
kodero: Do we need a `commit` here really?

tests/test_user_model.py
```diff
@@ -0,0 +1,31 @@
...
+    assert new_user.name == 'Ayobami' 
+
+  def test_count(self, new_user):
+    assert new_user.count()  == 1 
+
+  def test_get_all(self, new_user):
+    assert isinstance(new_user.get_all(), list)
+
+  def test_delete(self, new_user):
+    new_user.delete()
```
kodero: Remove the excess space after this.

purpose50:
Thanks for the review @Victoradukwu , The user cannot clone into an existing repo, So everytime you clone this branch(not pull from this branch into an existing repo) then you need to create a virtual env and run all necessary do all other neccessary setup

purpose50:
Thanks, @Victoradukwu, You can't clone into an existing repo. So every time you clone this branch(not pull from this branch into an existing repo) you need to create a virtual env and do any other necessary setup

.pytest_cache/v/cache/nodeids
```diff
@@ -0,0 +1,8 @@
+[
```
ibutiti: Shouldn't this file be ignored in .gitignore?

purpose50: Thanks, @ibutiti, will do that right away, I actually thought I ignored it.

api/models/model_operations.py
purpose50: Hi Kodero,  when I changed the method name to ```query```, I get a ```BaseQuery not callable``` error, so I figured everytime I called ```Model.query``` it was probably calling the query method defined in  ```db.Model``` which is probably not callable. to fix this issue I used ```querying``` as the method name instead of ```query``` which is a keyword for our model. I also improved the method so we can implement filtering on it. the test case for this asserts a couple of filtering

api/models/model_operations.py
purpose50: okay

api/models/model_operations.py
purpose50: On it

api/models/model_operations.py
purpose50: Thanks, @kodero , Yeah, I will do that right away.

api/models/model_operations.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def delete(self):
+        """
+        Delete a model instance.
+        """
+        db.session.delete(self)
+        db.session.commit()
+
+    @classmethod
+    def querying(cls):
```
kodero: Please change this to `_query`

.gitignore
```diff
@@ -2,5 +2,8 @@
 .vscode
 .DS_STORE
 __pycache__
+venv
+.pytest_cache
+.pytest_cache/v/cache
 .idea/
-activo-env/
\ No newline at end of file
+activo-env/
```
kodero: The project readme says virtual environment name is `venv`, so we don't need this

PR #10, opened 2018-06-01
#157962202 Create AssetCategory, Attribute and Asset models

GbengaOshinaga:


api/models/model_operations.py
```diff
@@ -1,4 +1,5 @@
 from .database import db
+from datetime import datetime
 from uuid import uuid4
 
 class ModelOperations(object):
```
purpose50: I was gonna ignore this earlier because I thought it probably was not in the scope of this task. but after Seun's review I think it's best for this to be done now. Since we are using a soft delete method then I think you should remove ```db.session.delete```. ```deleted``` should also be set to ```True``` whenever  a row in the database gets deleted, you should also set ```delete_by ``` to an hard-coded value for now.

the delete method in the model operations class should look something like this:

```
  def delete(self):
        """
        Delete a model instance.
        """
        self.deleted = True;
        self.deleted_at = datetime.utcnow()
        self.deleted_by = 'Operations Team'
        db.session.commit()

```

kodero: @GbengaOshinaga, @purpose50; Ideally, this should not change. The object of this change is to model the relationships between `Asset`, `AssetCategory` and any of its dynamic `Attributes`. Any changes on `ModelOperations` mixin will not only affect `delete`, but `get`, `_query` and `update` as well.

GbengaOshinaga: @kodero To gain clarity, do you mean this change is not supposed to be part of this PR, and should be removed as changes on this deleted method will affect other methods of the `ModelOperations` class?

kodero:
Yes. Precisely!


On Jun 4, 2018 12:05 PM, "Gbenga Oshinaga" <notifications@github.com> wrote:

*@GbengaOshinaga* commented on this pull request.
------------------------------

In api/models/model_operations.py
<https://github.com/andela/activo-api/pull/10#discussion_r192671920>:

> @@ -1,4 +1,5 @@
 from .database import db
+from datetime import datetime
 from uuid import uuid4

 class ModelOperations(object):

@kodero <https://github.com/kodero> To gain clarity, do you mean this
change is not supposed to be part of this PR, and should be removed as
changes on this deleted method will affect other methods of the
ModelOperations class?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/andela/activo-api/pull/10#discussion_r192671920>, or mute
the thread

GbengaOshinaga:
@kodero Done, changes have been removed

api/models/model_operations.py
kodero: @GbengaOshinaga, Yes. Precisely!

GbengaOshinaga:
@kodero The step has been added

PR #11, opened 2018-06-06
#157925463 fix bug with base model id attribute

Dheavyman:
#### What does this PR do?
- Fix issue with the base model id attribute which raises an error when the id is being assessed
- Remove UUID default value from the id attribute of the base model class
- Clone the repository and change directory into it
- Change to the branch with the command `git checkout -b bg-base-model-id-157925463`
- Create a virtual environment and install the packages using `pip install -r requirements.txt`
- Start the application and create a user using the user model class already defined
- The id for the user can be verified from the database to have been generated by the custom function and not from UUID
- The id field has been configured to be populated by a custom function and does not require the use of UUID for generating the values
[#157925463](https://www.pivotaltracker.com/story/show/157925463)
- N/A
#### Questions:
- N/A

### DIFF
--- a/api/models/base/base_model.py
+++ b/api/models/base/base_model.py
@@ -11,5 +11,5 @@ class BaseModel(db.Model, ModelOperations):
 
     __abstract__ = True
 
-    id = db.Column(db.String(36), primary_key=True, default=uuid.uuid4)
+    id = db.Column(db.String(36), primary_key=True)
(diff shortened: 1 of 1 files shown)


Dheavyman:
@ibutiti The custom id generator has already been incorporated into the app by @GbengaOshinaga. It does not need to be added from the base model class

ibutiti:
Oh, seen it. Code LGTM

PR #15, opened 2018-06-07
157876086 Endpoint to delete asset-category

purpose50:
#### What does this PR do?
delete asset categories

* create an endpoint to delete asset categories
* write test for the asset category deletion

* clone the branch
* create a new virtual env
* navigate to ```http://127.0.0.1:5000/api/v1/asset_categories/<string:id>``` on postman

* changed ```init_db``` function scope to ```module``` so the database will be dropped and recreated when every test module runs
* modify the delete method in ```model operations``` to implement a soft delete method

#157876086 An endpoint to delete asset-category

screen shot 2018-06-07 at 1 45 25 am
screen shot 2018-06-07 at 1 45 59 am

#### Questions:
N/A

### DIFF
--- a/api/models/asset_category.py
+++ b/api/models/asset_category.py
@@ -29,3 +29,4 @@ class AssetCategory(AuditableBaseModel):
 AssetCategory.assets_count = column_property(
     select([func.count(Asset.id)])
     .where(Asset.asset_category_id == AssetCategory.id))
+
--- a/api/models/model_operations.py
+++ b/api/models/model_operations.py
@@ -53,6 +53,8 @@ class ModelOperations(object):
         relationships = self.get_child_relationships()
         if delete_validator(relationships):
             self.deleted = True
+            if request.decoded_token:
+                self.deleted_by = request.decoded_token['UserInfo']['name']
             db.session.add(self)
             db.session.commit()
         else:
--- a/api/views/asset_category.py
+++ b/api/views/asset_category.py
@@ -83,6 +83,17 @@ class AssetCategoryStats(Resource):
             'data': data
         }
 
+    #         @token_required
+    # def delete(self, id):
+    #     """
+    #     Soft delete asset categories
+
+    #  single_category.delete()
+
+    #     return {
+    #         'status': 'success',
+    #         'message': 'category deleted successfully'
+
 
 @api.route('/asset-categories/<string:id>')
 class AssetCategoryListResource(Resource):
@@ -111,3 +122,23 @@ class AssetCategoryListResource(Resource):
                 single_category.attributes).data
            }
         }, 200
+
+    @token_required
+    def delete(self, id):
+        """
+        Soft delete asset categories
+        """
+        if not validate_id(id):
+            raise ValidationError(dict(message='Invalid id'), 400)
+
+        single_category = AssetCategory.get(id)
+        if not single_category or single_category.deleted:
+            raise ValidationError(dict(
+                message='Asset category not found'), 404)
+
+        single_category.delete()
+
+        return {
+            'status': 'success',
+            'message': 'category deleted successfully'
+        }
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@
 import os
 
 import pytest
+from flask import current_app, request
 
 from main import create_app
 from config import config
@@ -86,7 +87,8 @@ def new_asset_category_with_non_deleted_asset(app):
 
 
 @pytest.fixture(scope='module')
-def new_asset_category_with_deleted_asset(app):
+def new_asset_category_with_deleted_asset(app, request_ctx,
+                                          mock_request_obj_decoded_token):
     """Fixture for asset category with a deleted child asset."""
     asset_category = AssetCategory(name='Laptop')
     asset_category = asset_category.save()
@@ -115,3 +117,31 @@ def auth_header(generate_token=generate_token):
         'Accept': MIMETYPE
 
     }
+
+
+@pytest.fixture(scope='module')
+def request_ctx():
+    """
+    Setup a request client, this gets executed for each test module.
+
+    :param app: Pytest fixture
+    :return: Flask request client
+    """
+    ctx = current_app.test_request_context()
+    ctx.push()
+    yield ctx
+    ctx.pop()
+
+
+@pytest.fixture(scope='module')
+def mock_request_obj_decoded_token():
+    """
+    Mock decoded_token from request object
+    """
+    decoded_token = setattr(request, 'decoded_token', {
+        'UserInfo': {
+            'name': 'Ayobami'
+        }
+    })
+
+    return decoded_token
--- a/tests/test_asset_categories_endpoints.py
+++ b/tests/test_asset_categories_endpoints.py
@@ -1,4 +1,5 @@
 "Module for asset category endpoint test"
+import pytest
 
 from os import getenv
 
@@ -22,7 +23,8 @@ class TestAssetCategoriesEndpoints:
     """
 
     def test_asset_categories_stats_endpoint(self, client, new_asset_category,
-                                             init_db, auth_header):
+                                             init_db, auth_header, request_ctx,
+                                             mock_request_obj_decoded_token):
         """
         Should pass when getting asset categories stats
         """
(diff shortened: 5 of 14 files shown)


tests/test_asset_category_endpoint.py
```diff
@@ -0,0 +1,23 @@
+from flask import json, request
+from api.models.asset_category import AssetCategory
+import os
+
+
+class TestAssetCategoriesEndpoint:
+    def test_asset_categories_delete(self, client, init_db):
+        params = {
```
ibutiti: I notice you're using a lot of variables that I think are not necessary. You could pass the `name` kwarg to AssetCategory directly since it's just one.

Another one is `id` on line 13

api/models/model_operations.py
```diff
@@ -18,7 +18,7 @@ def update(self, **kwargs):
         """
         for field, value in kwargs.items():
             setattr(self, field, value)
-            db.session.commit()
+        db.session.commit()
```
ibutiti: Nice :)

GbengaOshinaga:
@purpose50 All PRs require a max of 5 commits, so you should squash yours to a maximum of at most 5

tests/test_asset_category_endpoint.py
purpose50: Thanks, @ibutiti, the variable ```name``` has been passed to AssetCategory as suggested, the same with the ```id```. the other variables such as ```client``` and ```init_db``` are test fixtures which are needed for this test to run properly. The fixtures were created in ```conftest.py``` file

tests/conftest.py
```diff
@@ -60,9 +61,24 @@ def new_asset_category(app):
...
+    _db.session.close()
     _db.drop_all()
 
+
+@pytest.fixture(scope='module')
+def auth_headers():
+    """
+    set up request header
+    """
+    myToken = os.getenv('TEST_TOKEN')
```
kodero: It would be preferable to pass `os.getenv('TEST_TOKEN')` directly without creating a local variable.

tests/test_asset_category_endpoint.py
```diff
@@ -0,0 +1,21 @@
...
+
+class TestAssetCategoryEndpoints:
+    """
+    Tests that a single asset category can be deleted
+    """
+    def test_delete_asset_category(self, client, init_db, auth_headers):
+        asset_category = AssetCategory(name='TestLaptop')
+        asset_category.save()
+
+        response = client.delete('/api/v1/asset_categories/{}'.format(
```
kodero: Abstract the base url `/api/v1` to an external config.

tests/test_asset_category_endpoint.py
```diff
@@ -0,0 +1,21 @@
...
+          asset_category.id), headers=auth_headers)
+          
+        response_json = json.loads(response.data.decode('utf-8'))
+  
+        assert response.status_code == 200
+        assert response_json['status'] == 'success'
+        assert response_json['message'] == 'category deleted successfully'
```
kodero: The test will still be valuable without this line.

tests/test_asset_category_endpoint.py
```diff
@@ -0,0 +1,21 @@
...
+    Tests that a single asset category can be deleted
+    """
+    def test_delete_asset_category(self, client, init_db, auth_headers):
+        asset_category = AssetCategory(name='TestLaptop')
+        asset_category.save()
+
+        response = client.delete('/api/v1/asset_categories/{}'.format(
+          asset_category.id), headers=auth_headers)
+          
+        response_json = json.loads(response.data.decode('utf-8'))
```
kodero: Let's have this(`'utf-8'`) as a constant somewhere and pull from there whenever we need it.

api/views/asset_category.py
```diff
@@ -30,3 +30,23 @@ def get(self):
             'status': 'success',
             'data': data
         }
+        
+
+@api.route('/asset_categories/<string:id>')
+class AssetCategoryResource(Resource):
+    """
+    include a docstring explaining what this view does
```
kodero: This needs to be fixed. ;-)

tests/conftest.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+import os
```
brayoh: let's have `os` import come first followed by `pytest` since `os` is a standard python library while `pytest` is a `third party` library, have a look at [python imports styleguide](https://www.python.org/dev/peps/pep-0008/#imports).

tests/test_asset_categories_endpoints.py
```diff
@@ -1,17 +1,6 @@
-from os import getenv
-from flask import json
+# from os import getenv
```
brayoh: Delete the following file since it's no longer needed

api/models/model_operations.py
```diff
@@ -25,7 +25,7 @@ def update(self, **kwargs):
         """
         for field, value in kwargs.items():
             setattr(self, field, value)
-        db.session.commit()
+            db.session.commit()
```
SEUNAGBEYE: @purpose50 This shouldn't be in a loop

purpose50: @SEUNAGBEYE that issue was already addressed in a different PR by Kodero and the team, We(The whole team) agreed on the call to create a bug fix for that instead of including the change in another task.

tests/test_asset_categories_endpoints.py
```diff
@@ -19,3 +19,28 @@ def test_asset_categories_stats_endpoint(self, client, new_asset_category, init_
...
+        assert response_json['status'] == 'success'
+
+    def test_delete_asset_category_not_found(self, client, init_db,
+                                             auth_header):
+        """
+         Tests that 404 is returned for a category that does not exist on
+         delete
+        """
+
+        response = client.delete('{}/asset-categories/2', headers=auth_header)
```
Dheavyman: @purpose50 You did not call the format method on the string to determine what goes into the curly braces.

requirements.txt
```diff
@@ -7,6 +7,7 @@ cffi==1.11.5
 click==6.7
 cryptography==2.2.2
 Flask==1.0.2
+flask-marshmallow==0.9.0
```
Dheavyman: This extension is not needed in the application. We are making use of the core marshmallow module.

api/models/asset_category.py
```diff
@@ -5,7 +5,6 @@
 from .database import db
 from .base.auditable_model import AuditableBaseModel
 
-
```
Dheavyman: Check if you are using proper spaces before a class according to PEP8

purpose50: @Dheavyman yes I am, but this file does not concern my task, I only removed an empty line flagged by pylint.

Dheavyman: you need to have two line spaces before the class is defined. you made modification that resulted in the changes showing on this PR

tests/conftest.py
```diff
@@ -1,5 +1,4 @@
 import os
-
```
Dheavyman: If you go through the link that Brian sent to us on import style guide https://www.python.org/dev/peps/pep-0008/#imports, you will see that there should be a blank line between each group of imports.

Dheavyman: If you go through the link that Brian sent to us on import style guide https://www.python.org/dev/peps/pep-0008/#imports you will see that there should be a blank line between each group of imports.

api/views/asset_category.py
```diff
@@ -30,3 +31,26 @@ def get(self):
...
+class AssetCategoryResource(Resource):
+    """
+    Resource class for deleting asset categories
+    """
+    @token_required
+    def delete(self, id):
+        """
+        soft delete asset categories
+        """
+        single_category = AssetCategory._query().filter_by(id=id, deleted=False).first()
```
Dheavyman: A suggestion, why not use the get method that is already defined in the model operations and then include if the single category deleted attribute is  True. Instead of using the _query method and filtering

purpose50: okay

api/models/model_operations.py
SEUNAGBEYE: @brayoh Should any one of us go ahead and create a story to fix the `db.session.commit` in the loop? Or you will create the story yourself?

tests/conftest.py
```diff
@@ -65,6 +65,12 @@ def new_asset_category(app):
     return asset_category
 
 
+"""
+the module scope is used here to prevent a test module data leaking into
```
GbengaOshinaga: Capitalize the first letter of your docstring

api/views/asset_category.py
```diff
@@ -79,3 +81,29 @@ def get(self):
             'status': 'success',
             'data': data
         }
+        
+
+@api.route('/asset-categories/<string:id>')
+class AssetCategoryListResource(Resource):
+    """
+    Resource class for deleting asset categories
+    """
```
brayoh: Add a blank line here

brayoh:
Also squash your commits into 1 and fix the merge conflicts

PR #17, opened 2018-06-07
#157767387 Add dynamic filter to support entity query

Victoradukwu:
#### What does this PR do?
- Enable dynamic filtering of entity query result
- Define a class that dynamically filters a query result
- Refactor the `_query` method of `model_operations` class to return filtered result
- Create a branch `ch-add-dynamic-filter-157767387` and pull from its remote counterpart
- Activate the python environment by running `venv/bin/activate`
- Run `pip install -r requirements.txt` to install all dependencies
- In your terminal run `flask run` to start the application.
- Use Postman, make a get request to the URL `http://127.0.0.1:5000/api/v1/asset-categories/stats`. This URL has no filter and should return all the asset categories in the database.
- Make filtered get requests to the URL in the format: `http://127.0.0.1:5000/api/v1/asset-categories/stats?where=x,y,z`. **x** is an attribute, such as `name`, of the AssetCategory class. **y** is comparator and **z** is the string value to compare.
- Possible values of **y** are: **eq**(equal to); **ne**(not equal to); **lt**(less than); **gt**(greater than); **le**(less than or equal to); **ge**(greater than or equal to); **like**(like). 
N/A
 - ch-add-dynamic-filter-to-support-query-operation-157767387
image

image

image

#### Questions:
N/A


api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,57 @@
...
+	def greater_than(column, value):
+		return column > value
+
+	def greater_or_equal(column, value):
+		return column >= value
+
+	def less_than(column, value):
+		return column < value
+
+	def less_or_equal(column, value):
```
SEUNAGBEYE: @Victoradukwu Is this a `static`, `class` or `abstract` method, I've never seen a method defined like this before in python

Victoradukwu: This is a static method. The @staticmethod decorator is only required if you intend calling the method from a class instance.

kodero:
Hey @Victoradukwu, please review the steps of `How should this be manually tested?`. Be explicit

api/models/model_operations.py
```diff
@@ -35,17 +35,17 @@ def delete(self):
         db.session.commit()
 
     @classmethod
-    def _query(cls):
+    def _query(cls, filter_condition):
         """
-        return all database entries
+        return filtered database entries
```
kodero: Use Docstrings to explain a little more of how the filtering works, especially, provide examples of the how the filter query params should be sent.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,41 @@
+from .filter_functions import (
+	like, is_equal, less_than, not_equal,
+	greater_than, less_or_equal, greater_or_equal
+							)
+
+
+class DynamicFilter:
+
+	def __init__(self, model):
```
kodero: Provide doc strings on what this class does.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,41 @@
...
+		'gt': greater_than,
+		'le': less_or_equal,
+		'ge': greater_or_equal,
+	}
+
+	def filter_query(self, raw_filters):
+		from api.middlewares.base_validator import ValidationError
+		result = self.query
+		if not raw_filters:
+			return 200, self.query.all()
```
kodero: We need to provide some limit and not query all items in the db.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,41 @@
...
+
+		for raw in raw_filters:
+			try:
+				key, op, value = raw.split(',', 3)
+			except ValueError:
+				raise ValidationError(dict(message='Invalid filter: %s' % raw))
```
kodero: It would important here to provide what the filter format should look like

api/utilities/dynamic_filter.py
Victoradukwu: I am of the opinion that the view function that makes use of this should impose the desired limit.

api/utilities/dynamic_filter.py
Victoradukwu: I have added that in docstring for the method

Victoradukwu:
@kodero @GbengaOshinaga @purpose50 @SEUNAGBEYE The requested changes have been applied

.flaskenv
```diff
@@ -1 +1 @@
-FLASK_APP=main
+FLASK_APP=manage
```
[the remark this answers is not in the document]
Victoradukwu: The app is created in the `manage.py` file. This is necessary so we can use the `flask run` command.

api/utilities/dynamic_filter.py
kodero: Yep. Seen.

api/utilities/dynamic_filter.py
SEUNAGBEYE: @Victoradukwu That's good.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,55 @@
...
+		"""
+		return filtered database entries.
+		An example of filter_condition is: User._query('name,like,john').
+		Apart from 'like', other comparators are
+		eq(equal to), ne(not equal to), lt(less than), le(less than or equal to)
+		gt(greater than), ge(greater than or equal to)
+		:param filter_condition:
+		:return: an array of filtered records
+		"""
+		from api.middlewares.base_validator import ValidationError
```
Dheavyman: Why not move this import to the top of the file.

api/utilities/dynamic_filter.py
Victoradukwu: This has been provided.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,41 @@
+from .filter_functions import (
```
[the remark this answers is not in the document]
Victoradukwu: Thank you

Victoradukwu:
@GbengaOshinaga @purpose50 @SEUNAGBEYE @kodero All requested changes have been effected on the PR.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,60 @@
...
+
+		result = self.query
+		if not raw_filters:
+			return self.query.all(), 200
+
+		for raw in raw_filters:
+			try:
+				key, op, value = raw.split(',', 3)
+			except ValueError:
+				raise ValidationError(dict(message='Invalid filter: %s' % raw))
```
kodero: Change this to `Invalid filter format` and provide an example format in the error message. The front-end team may not have time or even have access to the source code to know the exact format.

api/views/asset_category.py
```diff
@@ -12,15 +13,15 @@ class AssetCategoryStats(Resource):
     their corresponding asset counts
     """
 
-    @token_required
+    # @token_required
```
kodero: Why was this commented out?

api/views/asset_category.py
```diff
@@ -12,15 +13,15 @@ class AssetCategoryStats(Resource):
...
     def get(self):
         """
         Gets asset categories and the corresponding asset count
         """
-
-        asset_categories = AssetCategory._query().all()
+        filter_condition = request.args.getlist('where')
```
kodero: Am of the idea we let the `_query()` deal with getting any params it wants to use from the request object. 
So the line below should read something like;
`asset_categories = AssetCategory._query(request.args)`
Then inside `_query()` of model operations, we get the `where` param as 
`filter_condition = args.getlist('where')` and use it appropriately.

api/views/asset_category.py
Victoradukwu: I was supposed to uncomment it after testing. I wanted to be able to test without a token. I have uncommented that now.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,63 @@
...
+				.format(raw)))
+
+			column = getattr(self.model, key, None)
+			if not column:
+				raise ValidationError(dict(
+					message='Invalid filter column: %s' % key))
+
+			db_filter = self.mapper.get(op)
+			result = result.filter(db_filter(column, value))
+		return result, 200
```
Dheavyman: @Victoradukwu I think you should have the status code in the view function and not here. Let this function return just the filter result then you can do with it whatever you want at the view and return the necessary status code there

kodero:
Hello @Victoradukwu, please write tests for this filter as well. You can elect to do it after this is merged, or do them right in this PR.

Victoradukwu:
@kodero I would prefer we get this PR merged while I go ahead and work on the test as required.

api/utilities/dynamic_filter.py
```diff
@@ -0,0 +1,63 @@
...
+		'lt': less_than,
+		'ne': not_equal,
+		'gt': greater_than,
+		'le': less_or_equal,
+		'ge': greater_or_equal,
+	}
+
+	def filter_query(self, args):
+		"""
+		return filtered database entries.
```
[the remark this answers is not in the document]
Victoradukwu: This has been implemented

api/models/model_operations.py
```diff
@@ -35,17 +35,24 @@ def delete(self):
...
-    def _query(cls):
-        """
-        return all database entries
-        """
-        all_entries = cls.query
-        return all_entries
- 
+    def _query(cls, filter_condition):
+        """
+        return filtered database entries. It takes model class and
```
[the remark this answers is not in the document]
Victoradukwu: @kodero The requested changes have been applied

PR #18, opened 2018-06-12
#158285755 Generate jwt tokens for test environment

SEUNAGBEYE:
#### What does this PR do?
Generates valid `jwt tokens` when testing endpoints that require the client to be authenticated.

- Create a function that generates `jwt tokens`
- Encryption algorithm should be `RS256`
- Token generated by the function should be valid with verified with the `token_required` decorator

Clone the branch
Run `pip install -r requrements.txt`
set `FLASK_ENV = testing` and `API_BASE_URL_V1=/api/v1`
Run `pytest`

If you need to test endpoints that require the client to be `authenticated` pass the `auth_header`  `fixture` to the `test function` and assign `headers=auth_header`.

[#158285755](https://www.pivotaltracker.com/n/projects/2170023/stories/158285755)

N/A

#### Questions:
N/A

### DIFF
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -1,3 +1,2 @@
 from flask import Blueprint
-api_v1_base_url = '/api/v1'
-api_blueprint = Blueprint('api_blueprint', __name__, url_prefix=api_v1_base_url)
+api_blueprint = Blueprint('api_blueprint', __name__, url_prefix='/api/v1')
--- a/api/__init__.py
+++ b/api/__init__.py
@@ -1,2 +1,3 @@
 from flask import Blueprint
-api_blueprint = Blueprint('api_blueprint', __name__, url_prefix='/api/v1')
+api_v1_base_url = '/api/v1'
+api_blueprint = Blueprint('api_blueprint', __name__, url_prefix=api_v1_base_url)
--- a/api/utilities/constants.py
+++ b/api/utilities/constants.py
@@ -1,0 +1,19 @@
+INVALID_TOKEN_MSG = dict(
+    message="Authorization failed due to an Invalid token."
+)
+EXPIRED_TOKEN_MSG = dict(
+    message="Token expired. Please login to get a new token"
+)
+SIGNATURE_ERROR = dict(
+    message="Cannot verify the signature of the token provided as \
+    it was signed by a non matching private key"
+)
+SERVER_ERROR_MESSAGE = dict(
+    message="Authorization failed. Please contact support."
+)
+NO_BEARER_MSG = dict(
+    message="Bad request. The token should begin with the word 'Bearer'."
+)
+NO_TOKEN_MSG = dict(
+    message="Bad request. Header does not contain an authorization token."
+)
--- a/api/utilities/constants.py
+++ b/api/utilities/constants.py
@@ -1,4 +1,4 @@
 "Module for storing constants"
 
-UTF_8 = 'utf-8'
+CHARSET = 'utf-8'
 MIMETYPE = 'application/json'
--- a/api/utilities/constants.py
+++ b/api/utilities/constants.py
@@ -1,3 +1,5 @@
+"Module for storing constants"
+
 INVALID_TOKEN_MSG = dict(
     message="Authorization failed due to an Invalid token."
 )
@@ -17,3 +19,14 @@ NO_BEARER_MSG = dict(
 NO_TOKEN_MSG = dict(
     message="Bad request. Header does not contain an authorization token."
 )
+
+UTF_8 = 'utf-8'
+
+jwt_errors = {
+    'INVALID_TOKEN_MSG': INVALID_TOKEN_MSG,
+    'EXPIRED_TOKEN_MSG': EXPIRED_TOKEN_MSG,
+    'SIGNATURE_ERROR': SIGNATURE_ERROR,
+    'SERVER_ERROR_MESSAGE': SERVER_ERROR_MESSAGE,
+    'NO_BEARER_MSG': NO_BEARER_MSG,
+    'NO_TOKEN_MSG': NO_TOKEN_MSG,
+}
--- a/api/utilities/constants.py
+++ b/api/utilities/constants.py
@@ -1,32 +1,4 @@
 "Module for storing constants"
 
-INVALID_TOKEN_MSG = dict(
-    message="Authorization failed due to an Invalid token."
-)
-EXPIRED_TOKEN_MSG = dict(
-    message="Token expired. Please login to get a new token"
-)
-SIGNATURE_ERROR = dict(
-    message="Cannot verify the signature of the token provided as \
-    it was signed by a non matching private key"
-)
-SERVER_ERROR_MESSAGE = dict(
-    message="Authorization failed. Please contact support."
-)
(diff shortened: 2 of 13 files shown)


api/utilities/constants.py
```diff
@@ -0,0 +1,19 @@
+INVALID_TOKEN_MSG = dict(
```
ibutiti: We've got a messages module for this. You could put related ones in a dictionary appropriately named, such as `jwt_errors` and import a single object in other files.

SEUNAGBEYE: @ibutiti done, but I left the `INVALID_TOKEN_MSG` and similar constants in the test file the way they were for the sake of readability, else I will be doing `jwt_errors['INVALID_TOKEN_MSG']['message']` which I don't find readable.

Victoradukwu:
In the section **Any background context you want to provide?**, do we set `headers = auth_headers` or `headers = auth_header`?

SEUNAGBEYE:
@Victoradukwu it's `headers = auth_header`

api/utilities/constants.py
kodero: @SEUNAGBEYE, use the messages module that has been done by @ibutiti & @Dheavyman. You should find it in inside `api/utilities/messages/error_messages`.

api/__init__.py
```diff
@@ -1,2 +1,3 @@
 from flask import Blueprint
-api_blueprint = Blueprint('api_blueprint', __name__, url_prefix='/api/v1')
+api_v1_base_url = '/api/v1'
```
Dheavyman: This should be saved as a config variable according to our last review on this PR, we can have it as an environment config variable

Dheavyman: This should be saved as a config variable according to our last review on this PR, we can have it as an environment config variable

SEUNAGBEYE: @Dheavyman I'm on it.

api/utilities/messages/error_messages.py
```diff
@@ -13,6 +13,36 @@
...
+    message="Authorization failed. Please contact support."
+)
+NO_BEARER_MSG = dict(
+    message="Bad request. The token should begin with the word 'Bearer'."
+)
+NO_TOKEN_MSG = dict(
+    message="Bad request. Header does not contain an authorization token."
+)
+
+jwt_errors = {
```
ibutiti: Please pass the errors directly to the JWT errors to avoid unnecessary variables. You can put in the `#noqa` inline comment for string lines that go past the 79 character limit so the linter doesn't complain.

tests/test_token_required_middleware.py
```diff
@@ -0,0 +1,44 @@
+from os import getenv
+from flask import json
+from api.utilities.messages.error_messages import (
+    NO_BEARER_MSG,
```
ibutiti: Replace these with the `jwt_errors` dictionary and use accordingly

api/utilities/constants.py
```diff
@@ -0,0 +1,4 @@
+"Module for storing constants"
+
+UTF_8 = 'utf-8'
```
ibutiti: I think this should have another name... perhaps something like `CHARSET` so we can use another charset if needed and not have a confusing variable name. What's your opinion @kodero?

api/utilities/messages/error_messages.py
SEUNAGBEYE: @ibutiti  I did it this way to avoid doing my assertion like this  `assert response_json['message'] == jwt_errors['INVALID_TOKEN_MSG']['message']` , instead I'm doing my assertion like this `assert response_json['message'] == INVALID_TOKEN_MSG['message']`

ibutiti: Why are you storing the error as a dict?

ibutiti: I think to make it truly generic we should maintain the string and generate a dict when raising the ValidationError

api/utilities/constants.py
ibutiti: @kodero recommends calling it `CHARACTER_SET`

Dheavyman: @ibutiti When you say "not have a confusing variable name", if we want to use say ASCII what will the name be. I think using CHARACTER_SET does not uniquely identify each character set.

api/utilities/messages/error_messages.py
SEUNAGBEYE: @ibutiti at the moment we're generating a `dict` when raising `ValidatorError` or do you mean we should save the errors as `strings` and do this `ValidationError(dict(message=jwt_errors['INVALID_TOKEN']))`? If yes, this looks a little bit clumsy to me compared to this `ValidationError(jwt_errors['INVALID_TOKEN'])` or `ValidationError(INVALID_TOKEN)` when raising `ValidationError` in the `token_required`

ibutiti: I mean like `ValidationError(dict(message=jwt_errors['INVALID_TOKEN']))`. It may seem a little clumsy but it's clear that you're assigning the error to a dictionary with the key messages. This is also consistent with all the other error messages in the messages module.

SEUNAGBEYE: @ibutiti This is resolve

api/utilities/constants.py
ibutiti: I mean it in the sense that if we were to switch from utf-8 to ASCII and the variable name is UTF_8, that will cause confusion

ibutiti: If we're to have a unique variable name for each character set then we may as well hard code it like we'd earlier done @Dheavyman

SEUNAGBEYE: @ibutiti Can't there be a situation where we would be using both?

ibutiti: If it's like that I don't see the benefit of a constant.

Dheavyman: @kodero @ibutiti @SEUNAGBEYE If there is no intention of having any other character set then we can use the CHARACTER_SET as the name

SEUNAGBEYE: @Dheavyman We are not sure of that yet.

kodero: @SEUNAGBEYE @Dheavyman, the reason we asked to externalize this variable name is to avoid the inconvenience of having to replace every occurance of `utf-8` were the `character set` ever need to be changed. Its therefore clear to better represent this `variable` as `CHARACTER_SET`.

tests/test_token_required_middleware.py
```diff
@@ -0,0 +1,40 @@
+from os import getenv
+from flask import json
+from api.utilities.messages.error_messages import jwt_errors
```
brayoh: it's usually advised to group module imports in the following order

```

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.
```
have a look at the following `pep8` styleguide for more info [imports styleguide](https://www.python.org/dev/peps/pep-0008/#imports)

brayoh:
@SEUNAGBEYE also kindly remember to `squash` your commits once you are done.

SEUNAGBEYE:
@brayoh PR fixed

brayoh:
@SEUNAGBEYE kindly `squash` the commits then update the ticket status on `PT`

PR #20, opened 2018-06-14
157876050 Endpoint to get one asset-category 

purpose50:
#### What does this PR do?
get one asset category

* create an endpoint to get asset category
* write a test for getting one category

* pull the branch
* navigate to ```http://127.0.0.1:5000/api/v1/asset_categories/<string:id>``` on postman

N/A

#157876050 An endpoint to get one asset category

screen shot 2018-06-14 at 12 51 15 pm

screen shot 2018-06-14 at 12 50 30 pm

#### Questions:
N/A

### DIFF
--- a/api/views/asset_category.py
+++ b/api/views/asset_category.py
@@ -1,3 +1,6 @@
+"""Module for asset category resource."""
+import re
+
 from flask_restplus import Resource
 from flask import request, jsonify
 
@@ -6,6 +9,8 @@ from main import api
 from api.middlewares.token_required import token_required
 from api.utilities.model_serializers.asset_category import AssetCategorySchema
 from api.utilities.model_serializers.attribute import AttributeSchema
+from api.middlewares.base_validator import ValidationError
+from api.utilities.validators.validate_id import validate_id
 
 
 @api.route('/asset-categories')
@@ -53,8 +58,6 @@ class AssetCategoryResource(Resource):
         return response
 
 
-
-
 @api.route('/asset-categories/stats')
 class AssetCategoryStats(Resource):
     """
@@ -79,3 +82,32 @@ class AssetCategoryStats(Resource):
             'status': 'success',
             'data': data
         }
+
+
+@api.route('/asset-categories/<string:id>')
+class AssetCategoryListResource(Resource):
+    @token_required
+    def get(self, id):
+        """
+        Get a single asset category
+        """
+        if not validate_id(id):
+            raise ValidationError(dict(message='Invalid id'), 400)
+
+        single_category = AssetCategory.get(id)
+        if not single_category or single_category.deleted:
+            raise ValidationError(dict(
+                message='Asset category not found'), 404)
+
+        asset_category_schema = AssetCategorySchema()
+        attributes_schema = AttributeSchema(
+            many=True, exclude=['choices', 'id', 'deleted'])
+
+        return {
+           'status': 'success',
+           'data': {
+            'name': single_category.name,
+            'customAttributes': attributes_schema.dump(
+                single_category.attributes).data
+           }
+        }, 200
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+"""Module for setting up fixtures for testing"""
 import os
 
 import pytest
@@ -66,7 +67,13 @@ def new_asset_category(app):
 
 @pytest.fixture(scope='module')
 def new_asset_category_with_non_deleted_asset(app):
-    """Fixture for asset category with a non deleted child asset."""
+    """
+    The module scope is used here to prevent a test module data leaking into
+    another.
+
+    Fixture for asset category with a non deleted child asset.
+    """
+
     asset_category = AssetCategory(name='Laptop')
     asset_category = asset_category.save()
 
--- a/tests/test_asset_categories_endpoints.py
+++ b/tests/test_asset_categories_endpoints.py
@@ -2,8 +2,9 @@
 
 from os import getenv
 
-from flask import json
+from flask import json, request
 
+from api.models.asset_category import AssetCategory
 from api.utilities.constants import CHARSET
 from .mocks.asset_category import (
     valid_asset_category_data,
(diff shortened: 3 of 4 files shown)


api/views/asset_category.py
```diff
@@ -30,3 +33,25 @@ def get(self):
...
             'data': data
         }
+
+
+@api.route('/asset-categories/<string:id>')
+class AssetCategoryResource(Resource):
+    @token_required
+    def get(self, id):
+        """
+        get a single asset categories
```
ibutiti: Should be `category` not `categories`. Also capitalise first letter.

api/views/asset_category.py
```diff
@@ -30,3 +33,25 @@ def get(self):
...
+        """
+        single_category = AssetCategory.get(id)
+        if not single_category or single_category.deleted is True:
```
ibutiti: The `is True` is unnecessary.

api/views/asset_category.py
```diff
@@ -30,3 +33,25 @@ def get(self):
...
+            raise ValidationError(dict(message='Asset category not found'), 404)
+
+        asset_category_schema = AssetCategorySchema()
+        attributes_schema = AttributeSchema(
+            many=True, exclude=['choices', 'id', 'deleted'])
+
+        return {
```
ibutiti: You need to serialize the entire object to json. Use jsonify.

tests/conftest.py
```diff
@@ -1,7 +1,7 @@
 import os
-
 import pytest
 
+
```
ibutiti: Please refer to the import statement arrangement recommended by @brayoh

ibutiti: Also, docstrings

api/views/asset_category.py
```diff
@@ -4,6 +4,9 @@
 from api.models import AssetCategory
 from main import api
 from api.middlewares.token_required import token_required
+from api.middlewares.base_validator import ValidationError
```
ibutiti: Docstrings

tests/test_asset_categories_endpoints.py
```diff
@@ -19,3 +19,31 @@ def test_asset_categories_stats_endpoint(self, client, new_asset_category, init_
...
+
+        response = client.get('{}/asset-categories/{}'.format(
+          api_v1_base_url, asset_category.id), headers=auth_header)
+
+        response_json = json.loads(response.data.decode(CHARSET))
+
+        assert response.status_code == 200
+        assert response_json['name'] == 'TestLaptop'
+        assert type(response_json['customAttributes']) == list
+
```
ibutiti: Also test for edge case of symbols, invalid strings etc as the id argument supplied in url so we see if the validation i'd recommended catches them.

api/views/asset_category.py
```diff
@@ -30,3 +33,25 @@ def get(self):
...
+
+
+@api.route('/asset-categories/<string:id>')
+class AssetCategoryResource(Resource):
+    @token_required
+    def get(self, id):
+        """
+        get a single asset categories
+        """
+        single_category = AssetCategory.get(id)
```
ibutiti: Since we know how the ids look, we can validate the id before hitting the db with a query.

api/views/asset_category.py
purpose50: Thanks @ibutiti  on it

api/views/asset_category.py
purpose50: Yes, thought I removed it already, on it.

api/views/asset_category.py
purpose50: Thanks @ibutiti , the reason why I didn't validate the `id` is that whatever the user supply as the `id`, as far as it is not in the id field of the table it will return `asset category not found` and the application will not break.

I also thought of what I should be validating for since the `id` we are expecting is a `string`, even if the user supplies an `integer`, the application won't break and the user will get an error that the `asset category is not found`. nevertheless, I'll try and `validate the id` to ensure it is a `string` so we don't leave any stone untouched.

purpose50: Hi @ibutiti made a new discovery. because the endpoint already defined the type of the id, it automatically converts it to  that type..for example:
this is the endpoint  `/asset-categories/<string:id>`. from the endpoint we are obviously expecting the id to be a string, so any id supplied from postman is automatically converted to a string..so all  of these will be automatically converted to a string:
`/asset-categories/1`
`/asset-categories/222`
`/asset-categories/@@@@`
`/asset-categories/0928ujj`
`/asset-categories/%%%%%`

now if the `id` suplied is not found in the database then the user gets a response that the `asset category is not found`, it doesnt break the app, it just returns a message that the `asset category with that id is not found` even though the `id` is just a set of random characters

brayoh:
@purpose50 Kindly `squash` your commits into 1

api/views/asset_category.py
Dheavyman: @ibutiti I believe flask-restplus has already converted the object to json response with the return. The header already has the application/json added too. From flask-restplus doc, "Flask-RESTPlus understands multiple kinds of return values from view methods. Similar to Flask, you can return any iterable and it will be converted into a response, including raw Flask response objects."

api/views/asset_category.py
ibutiti: Agreed @purpose50. However, since we know that the push_id uses these characters: ```PUSH_CHARS = ('-0123456789'
                  'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
                  '_abcdefghijklmnopqrstuvwxyz')```
we can do a simple regex to check that the characters in the id only appear here. This operation will be a lot cheaper than a DB query to let us know that the asset category was not found.

api/views/asset_category.py
ibutiti: ok @Dheavyman

tests/test_asset_categories_endpoints.py
```diff
@@ -19,3 +20,31 @@ def test_asset_categories_stats_endpoint(self, client, new_asset_category, init_
         assert len(response_json['data']) is not 0
         assert response_json['data'][0]['name'] == 'Laptop'
         new_asset_category.delete()
+
+    """
+    Tests that a single asset category can be retrieved
```
ibutiti: Please put this docstring inside the test

purpose50: sure, thanks

api/views/asset_category.py
SEUNAGBEYE: @ibutiti I concur, but we still need to verify if the item is in the `db`

```
if not (re.match('-[a-zA-Z0-9]', id) and Asset_category.get(id)):
                        raise Validation({'message': 'Asset Category not found'})
```

SEUNAGBEYE: @purpose50 I suggest you put that in a function somewhere in the `utilities` folder or somewhere else because we would be validating the `ids` in several places and it won't be idea repeating the validation

api/views/asset_category.py
```diff
@@ -30,3 +36,29 @@ def get(self):
...
+
+
+@api.route('/asset-categories/<string:id>')
+class AssetCategoryResource(Resource):
+    @token_required
+    def get(self, id):
+        """
+        get a single asset category
+        """
+        if not (re.match('-[a-zA-Z0-9]', id)):
```
Dheavyman: Please abstract this to a function in utilities folder so that it can be used in other places required.

api/views/asset_category.py
```diff
@@ -30,3 +37,29 @@ def get(self):
...
+
+        asset_category_schema = AssetCategorySchema()
+        attributes_schema = AttributeSchema(
+            many=True, exclude=['choices', 'id', 'deleted'])
+
+        return {
+           'name': single_category.name,
+           'customAttributes': attributes_schema.dump(
+               single_category.attributes).data
+        }, 200
```
Dheavyman: The response object should have status and data properties, with the data holding the information as described above. The status should either be success or error depending on the result of the request.

tests/conftest.py
```diff
@@ -64,6 +65,12 @@ def new_asset_category(app):
     return asset_category
 
 
+"""
+The module scope is used here to prevent a test module data leaking into
+another
+"""
+
+
```
brayoh: I think it would be better if we have this doc string inside the test method instead

PR #25, opened 2018-06-21
#158354496 get asset category attributes

Dheavyman:
#### What does this PR do?
- Have endpoint to get asset category attributes
- As a developer, I can get a list of the dynamic attributes for the provided asset category
- Create a local copy with the following commands: `git fetch origin ft-asset-categories-list-158320913:ft-asset-categories-list-158320913`, then `git checkout ft-asset-categories-list-158320913`
- Start the application `python manage.py`
- Send a GET request to this endpoint: `localhost:5000/api/v1/asset-categories/id/attributes`
- Run the tests with the command `pytest tests/test_attributes`
- N/A
[#158354496](https://www.pivotaltracker.com/story/show/158354496)
screen shot 2018-06-21 at 1 13 41 pm
screen shot 2018-06-21 at 1 14 13 pm
screen shot 2018-06-21 at 1 14 27 pm

#### Questions:
- N/A

### DIFF
--- a/api/utilities/validators/validate_id.py
+++ b/api/utilities/validators/validate_id.py
@@ -1,8 +1,25 @@
 """Module to validate resource id from url parameters"""
 
 import re
+from functools import wraps
+
+from api.middlewares.base_validator import ValidationError
 
 
-def validate_id(id):
-    """Validate id string"""
+def is_valid_id(id):
+    """Check if id is valid"""
     return re.match('^[-a-zA-Z0-9_]*$', id)
+
+
+def validate_id(function):
+    """Decorator function for views to validate id"""
+
+    @wraps(function)
+    def decorated_function(*args, **kwargs):
+        if not is_valid_id(kwargs.get('id', None)):
+            raise ValidationError({
+                'status': 'error',
+                'message': 'Invalid id in parameter'
+            }, 400)
+        return function(*args, **kwargs)
+    return decorated_function
--- a/api/views/asset_category_attributes.py
+++ b/api/views/asset_category_attributes.py
@@ -1,0 +1,40 @@
+"""Module for asset category attribute views"""
+
+from flask_restplus import Resource
+
+from main import api
+from ..models.asset_category import AssetCategory
+from ..utilities.model_serializers.attribute import AttributeSchema
+from ..middlewares.token_required import token_required
+from ..middlewares.base_validator import ValidationError
+from api.utilities.validators.validate_id import validate_id
+
+
+@api.route('/asset-categories/<string:id>/attributes')
+class AssetCategoryAttributes(Resource):
+    """Resource class for asset categories attributes"""
+
+    @token_required
+    @validate_id
+    def get(self, id):
+        """Get attributes of an asset category"""
+
+        asset_category = AssetCategory.get(id)
+        if not asset_category or asset_category.deleted:
+            raise ValidationError({
+                'status': 'error',
+                'message': 'Asset category not found'
+            }, 404)
+
+        attributes_schema = AttributeSchema(many=True, exclude=['deleted'])
+        attributes = attributes_schema.dump(asset_category.attributes
+                                            .filter_by(deleted=False)).data
+
+        return {
+            'status': 'success',
+            'message': 'Asset category attributes retrieved',
+            'data': {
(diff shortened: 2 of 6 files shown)


tests/test_attributes.py
```diff
@@ -0,0 +1,71 @@
...
+
+from flask import json
+
+from api.models.asset_category import AssetCategory
+from api.models.attribute import Attribute
+from api.utilities.constants import CHARSET
+
+api_base_url_v1 = getenv('API_BASE_URL_V1')
+
+
```
brayoh: remove the extra line

Dheavyman: @brayoh This line is necessary according to PEP8, two spaces are required before declaring a class.

brayoh: @Dheavyman I just realized it's before the class, sorry about that.

brayoh:
@Dheavyman get more reviews on this so that we can merge

brayoh:
also fix the merge conflicts

Dheavyman:
Merge conflict fixed

api/views/asset_category_attributes.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @token_required
+    def get(self, id):
+        """Get attributes of an asset category"""
+
+        if not validate_id(id):
+            raise ValidationError({
+                'status': 'error',
+                'message': 'Invalid asset category id'
+            }, 400)
```
SEUNAGBEYE: Good, but Instead of the` validate_id` function returning true or false, can we have some like this

```
def validate_id(id):
        if does not match regex exception:
               raise ValidationError({
               'status': 'error',
               'message': 'Invalid asset category id'
              }, 400)
```

then we can call `validate_id(id)` wherever we need it, this way we won't raise `ValidationError` in every place we need to validate the `id`. We do it one place and we're good, if we need to change something about the error returned we just change it one place and we're fine too

brayoh:
@Dheavyman Kindly squash your commits into 1

api/views/asset_category_attributes.py
Dheavyman: @SEUNAGBEYE The essence of this validate_id function is to check if an id is valid or not by returning True or False and affords the developer the flexibility to perform any other logic which includes but might not be limited to raising a ValidationError.

SEUNAGBEYE: @Dheavyman Oh I see, but we could still have a separate function that calls the `validate_id` function and return the appropriate errors so we don't have to do this ` raise ValidationError({
               'status': 'error',
               'message': 'Invalid asset category id'
              }, 400)` in any place we want to validate the `id`. It could be a `decorator` that we can use to decorate any function that want to validate the `id`. This way we don't have any code for id validation in our `views` and we could still use the `validate_id` seperately wherever we need to validate `id`, but we don't want to raise an `Exception`.

api/utilities/validators/validate_id.py
```diff
@@ -1,8 +1,25 @@
...
+    """Check if id is valid"""
     return re.match('^[-a-zA-Z0-9_]*$', id)
+
+
+def validate_id(function):
+    """Decorator function for views to validate id"""
+
+    @wraps(function)
+    def decorated_function(*args, **kwargs):
+        if not is_valid_id(kwargs['id']):
```
SEUNAGBEYE: @Dheavyman You deserve some accolades, nice job, one more thing though, could you please use `kwargs.get('id')(Does not throw KeyError if the key doesn't exist)` instead of `kwargs[id](Throws KeyError if the key doesn't exist)`

Dheavyman: Thanks for pointing that out, done.

SEUNAGBEYE: You welcome @Dheavyman.

api/utilities/validators/validate_id.py
```diff
@@ -1,8 +1,25 @@
...
+
+def is_valid_id(id):
+    """Check if id is valid"""
     return re.match('^[-a-zA-Z0-9_]*$', id)
+
+
+def validate_id(function):
+    """Decorator function for views to validate id"""
+
+    @wraps(function)
```
ibutiti: Nice

PR #31, opened 2018-06-26
#158592755 Filter Query by Deleted field

GbengaOshinaga:


api/views/asset_category.py
```diff
@@ -148,10 +148,9 @@ def get(self):
         if request.args:
             asset_categories = AssetCategory._query(request.args)
         else:
-            asset_categories = AssetCategory._query().filter_by(deleted=False)\
-                .with_entities(AssetCategory.id, AssetCategory.name, AssetCategory.assets_count)
+            asset_categories = AssetCategory._query().filter_by(deleted=False)
```
ibutiti: Can we put this as a default filter when request.args is None so we execute the filter from the db instead of filtering them after the query?

GbengaOshinaga:
@ibutiti Please check and confirm if changes are appropriate

api/utilities/dynamic_filter.py
```diff
@@ -59,5 +59,5 @@ def filter_query(self, args):
 					message='Invalid filter column: %s' % key))
 
 			db_filter = self.mapper.get(op)
-			result = result.filter(db_filter(column, value))
+			result = result.filter(db_filter(column, value)).filter_by(deleted=False)
```
ibutiti: Please have a look at the `_and` method to join the 2 conditions inside the `result.filter()`

GbengaOshinaga: @ibutiti That `and_` method doesn't seem to be able to work here, as it doesn't accept keyword arguments. Normally both `filter` and `filter_by` takes two arguments which it converts to `and_`, but `filter` takes only positional arguments while `filter_by` takes only keyword arguments, so it seems the only solution is to chain the `filter` and `filter_by`

PR #32, opened 2018-06-27
158592988 Create method to validate parameters

purpose50:
#### What does this PR do?
validate filtering parameters

* create a method to validate filtering parameters
* write tests for the method

* pull the branch
* test this on postman ```http://127.0.0.1:5000/api/v1/asset-categories/stats?where=created_at,eq,218-06-19 12:06:43.339809```
* run ```pytest ``` to run the test

The Dynamic filter class is an already existing class which used tabs for indentation. The team is using spaces for indentation and on account of that, there was indentation error using spaces in this file. to fix the indentation error I had to re-implement the file to use spaces instead of tabs.

#158592988 Implement helper methods to handle validation of parameter

screen shot 2018-06-27 at 12 46 21 pm

screen shot 2018-06-27 at 12 47 01 pm

screen shot 2018-06-27 at 12 48 43 pm

screen shot 2018-06-27 at 12 47 58 pm

#### Questions:
N/A


api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,122 @@
...
+        """
+
+        if op not in cls.mapper:
+            raise ValidationError(dict(message='''invalid operator, valid \
+operators are 'like, eq, lt, ne, gt, le, ge' '''))
+
+        try:
+            if key in cls.fields:
+                if key == 'created_at' or key == 'updated_at' or key == 'deleted_at':
+                    dt = datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f")
```
Dheavyman: dt as a variable name is not descriptive enough and I will suggest you go look for a more descriptive name if you still intend to define that variable. I said if because the variable seems to not be used anywhere.

GbengaOshinaga: +1

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,122 @@
...
+
+        if op not in cls.mapper:
+            raise ValidationError(dict(message='''invalid operator, valid \
+operators are 'like, eq, lt, ne, gt, le, ge' '''))
+
+        try:
+            if key in cls.fields:
+                if key == 'created_at' or key == 'updated_at' or key == 'deleted_at':
+                    datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f")
+                elif key == 'deleted' and value != 'true' and value != 'false':
```
Dheavyman: This is nice but I think a more readable way of implementing this will be `elif key == 'deleted and value not in ('true', 'false')`

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,122 @@
...
+        database query which therefore improves perfomance
+        """
+
+        if op not in cls.mapper:
+            raise ValidationError(dict(message='''invalid operator, valid \
+operators are 'like, eq, lt, ne, gt, le, ge' '''))
+
+        try:
+            if key in cls.fields:
+                if key == 'created_at' or key == 'updated_at' or key == 'deleted_at':
```
Dheavyman: You can also do the same as I suggested below instead of having the long chain of `or`

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,121 @@
...
+        http://127.0.0.1:5000/api/v1/asset-categories/stats?where=
+        deleted,invalid_operator,true
+        
+        The essence of this is to validate parameters before making a call to
+        the database, if the parameters are invalid then we dont make any
+        database query which therefore improves perfomance
+        """
+
+        if op not in self.mapper:
+            raise ValidationError(dict(message='''invalid operator, valid \
```
ibutiti: Lets put all our error messages in the error messages folder and call them through the dict like we've been doing.

purpose50: Sure thing @ibutiti

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,121 @@
...
+        raw_filters = args.getlist('where')
+        result = self.query
+        if not raw_filters:
+            return self.query.all()
+
+        for raw in raw_filters:
+            try:
+                key, op, value = raw.split(',', 3)
+            except ValueError:
+                raise ValidationError(dict(message='Invalid filter format:'
```
ibutiti: Please put this in the error messages file

purpose50: Hi @ibutiti I didn't want to touch that part of the code because it's not part of the changes I Introduced, it's flagged green because I changed the indentation, in the PR story above I mentioned about that in the `any background context you want to provide section`

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,120 @@
...
+        'lt': less_than,
+        'ne': not_equal,
+        'gt': greater_than,
+        'le': less_or_equal,
+        'ge': greater_or_equal,
+    }
+
+    def validate_query(self, key, value, op):
+        """
+        Validate parameters passed into any endpoint for filtering.
```
SEUNAGBEYE: could you please change the `into` to `to`

purpose50: 👍

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,120 @@
...
+        """
+        fields = ['deleted', 'created_at', 'updated_at', 'deleted_at']
+
+        if op not in self.mapper:
+            raise ValidationError(dict(
+                message=filter_errors['INVALID_OPERATOR']))
+
+        try:
+            if key in fields:
+                if key in ('created_at', 'updated_at', 'deleted_at'):
```
SEUNAGBEYE: @purpose50 I suggest you use this `if key in fields[1:]`(Do not forget to comment this so we know what you were trying to achieve), this way if you want to change any value in the `tuple` you just change it in the `fields` array. What do you think?

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,120 @@
...
+        database query which therefore improves perfomance
+        """
+        fields = ['deleted', 'created_at', 'updated_at', 'deleted_at']
+
+        if op not in self.mapper:
+            raise ValidationError(dict(
+                message=filter_errors['INVALID_OPERATOR']))
+
+        try:
+            if key in fields:
```
SEUNAGBEYE: @purpose50 What happens if `key is not in fields`?

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,120 @@
...
+        created_at,eq,invalid_date
+
+        Invalid operator example
+        ========================
+        http://127.0.0.1:5000/api/v1/asset-categories/stats?where=
+        deleted,invalid_operator,true
+        
+        The essence of this is to validate parameters before making a call to
+        the database, if the parameters are invalid then we dont make any
+        database query which therefore improves perfomance
```
SEUNAGBEYE: @purpose50 Good job on the documentation, you literally told me what this code is supposed to do. You deserve some accolades

purpose50: @SEUNAGBEYE We are only validating for fields that have to be in a certain format, E.g Date has to be in a particular format, so we are validating the query to check if the date provided is in the correct format, the same goes for deleted which must be a boolean.

api/utilities/dynamic_filter.py
purpose50: @SEUNAGBEYE We are only validating for fields that have to be in a certain format, E.g Date has to be in a particular format, so we are validating the query to check if the date provided is in the correct format, the same goes for deleted which must be a boolean.

api/utilities/dynamic_filter.py
Dheavyman: @purpose50 @SEUNAGBEYE Nice idea but if we are to go by your idea then he should make sure he adds a good comment specifying what we actually need from the field array at this point. This is necessary for future maintainability because if new values are added to the fields or the order is rearranged then this will break the code. What's your thought?

SEUNAGBEYE: @Dheavyman I agree with you, the array may be `re-arrange`, so I suggest we live it this way. @purpose50

api/utilities/dynamic_filter.py
brayoh: nice!

tests/test_dynamic_filter/test_dynamic_filter_query.py
```diff
@@ -0,0 +1,90 @@
...
+        :param like, eq, lt, ne, gt, le, ge
+
+        any operator other than those should throw an invalid operator error
+        and return an error in this format
+
+        :return {
+            "message": "invalid operator, valid operators are 'like, eq, lt, 
+                        ne, gt, le, ge",
+            "status": "error",
+            "status_code": 400
```
brayoh: remember to remove the status code in the payload

tests/test_dynamic_filter/test_dynamic_filter_query.py
```diff
@@ -0,0 +1,90 @@
...
+    def test_invalid_delete_attr_in_query(self, client, init_db, auth_header):
+        """
+        The delete attribute for filtering should be a boolean value,
+        anything other than a boolean should throw an error and return a
+        response in this format
+
+        :return {
+            "message": "deleted should be of type bool",
+            "status": "error",
+            "status_code": 400
```
brayoh: same here

tests/test_dynamic_filter/test_dynamic_filter_query.py
```diff
@@ -0,0 +1,90 @@
...
+    def test_invalid_date_attr_in_query(self, client, init_db, auth_header):
+        """
+        Any date attribute passed for filtering should be a valid date format,
+        anything other than that should throw an error and return a
+        response in this format
+
+        :return {
+            "message": "Invalid datetime format",
+            "status": "error",
+            "status_code": 400
```
brayoh: and here

brayoh:
@purpose50 do you have the postman documentation for this?

brayoh:
@purpose50 kindly fix the merge conflicts

api/utilities/dynamic_filter.py
```diff
@@ -1,63 +1,120 @@
...
+    def __init__(self, model):
+        """
+        Constructor to initialize an instance of the class.
+        :param model: the model class that will be using the methods of
+        this class
+        """
+        self.model = model
+        self.query = model.query
+
+    mapper = {
```
brayoh: How would I add 2 conditions in one query since there isn't an and operator? eg: `deleted,eq,true,and,created_at,eq,<timestamp>`

purpose50: okay, so basically, my task is to `validate the query`, I had to change the `indentation` of the whole file that is why it flagged them as my changes, the only part of the code I included is the `validate_query method` and the `tests`. but to answer your question, you can do something like this: `http://127.0.0.1:5000/api/v1/asset-categories?where=created_at,eq,2018-06-19 12:06:43.339809&where=deleted,eq,true`. and it is already implemented before this task

brayoh: Ooh! cool! makes sense, I thought they are your changes

PR #37, opened 2018-07-04
#158763804 refactor seeders

Dheavyman:
#### What does this PR do?
- Refactor seeders module to reflect changes in the models
- The seeders should be refactored to account for changes made to the models. It should also be separated into modules for better maintainability
- Create a local copy with the following commands: `git fetch origin ch-refactor-seeders-158763804:ch-refactor-seeders-158763804`, then `git checkout ch-refactor-seeders-158763804`
- Export the flask application with the command `export FLASK_APP=manage.py`
- Seed the database with the command `flask seed_database`
- N/A
[#158763804](https://www.pivotaltracker.com/story/show/158763804)
- N/A
#### Questions:
- N/A


brayoh:
LGTM.

seeders/seed_asset.py
```diff
@@ -0,0 +1,57 @@
...
+
+from api.models import AssetCategory, Asset
+
+
+def seed_asset():
+    # Apple tv assets
+    apple_tv_category = AssetCategory._query().filter_by(
+        name='Apple TV').first()
+    apple_tv_assets = [
+        Asset(tag='AND/345/EWRD', serial='GRGR334TGD',
```
ibutiti: List comprehensions can do this without the duplicated code. Helps with maintainability.

seeders/seed_asset.py
```diff
@@ -0,0 +1,57 @@
...
+        Asset(tag='AND/K32/003', serial='KDJS43JN432LP',
+              asset_category_id=laptops_category.id),
+        Asset(tag='AND/K32/004', serial='KDJS43JN432LP',
+              asset_category_id=laptops_category.id),
+        Asset(tag='AND/K32/005', serial='KDJS43JN432LP',
+              asset_category_id=laptops_category.id)
+    ]
+
+    for asset in laptops_asset:
+        asset.save()
```
ibutiti: You can do `db.session.add(asset)` then call `db.session.commit()` outside the loop

seeders/seed_asset.py
Dheavyman: @ibutiti Please, can you throw more light on How list comprehension should be utilized here.

ibutiti: You could place the attributes in a tuple of the structure `(tag, setial, category_id)` like `attribute_tuple =(('AND/345/EWRD', 'GRGR334TGD', apple_tv_category.id), (next set...))` then use a list comprehension to do the asset instances like so:
`[Asset(tag=attr[0], serial=attr[1], asset_category_id=attr[2]) for attr in attribute_tuple]`

Dheavyman: Yeah, I just found out that having the data separately will be more maintainable because the data can still be modified to contain custom attributes later. Thanks

Dheavyman: Yeah, I think abstracting the data makes it more maintainable but I will because the tuple does not immediately communicate what the values are, I will make use of objects instead. This also makes it more readable.

seeders/seed_asset.py
Dheavyman: @ibutiti Nice, I actually thought of doing that but it will require I import db object here to have access to the session. The asset instances already inherited the model operations methods which give us the ability to save the instances. I believe this is for optimization on database level but do we really need such optimization in this case?

seeders/__init__.py
```diff
@@ -1,56 +1,7 @@
...
-
-    chairs = AssetCategory(name='chairs')
-    chairs_attributes = [
-        Attribute(label='type', is_required=True,
-                  input_control='input-field', choices='multiple choices')
-    ]
-    for attribute in chairs_attributes:
-        chairs.attributes.append(attribute)
-    chairs.save()
+def seed_db():
```
SEUNAGBEYE: Nice

Dheavyman:
@Ibutiti Please find time to review the followup comments and my changes

PR #38, opened 2018-07-04
#158792176 Refactor Error Message

GbengaOshinaga:
screen shot 2018-07-04 at 10 00 06 pm


api/models/model_operations.py
```diff
@@ -58,9 +58,13 @@ def delete(self):
             db.session.add(self)
             db.session.commit()
         else:
+            relationship_names = []
+            for relationship in relationships:
+                relationship_names.append(relationship.first().__class__.__name__ + '(s)')
```
ibutiti: Please use f strings

api/models/asset_category.py
```diff
@@ -20,7 +20,7 @@ class AssetCategory(AuditableBaseModel):
                                  cascade='save-update, delete', lazy='dynamic')
 
     def get_child_relationships(self):
-        return self.assets, self.attributes
+        return (self.assets,)
```
ibutiti: We need to check for non-deleted child attributes too. Is there a reason you've removed this?

SEUNAGBEYE: @ibutiti I believe an asset category should be deletable even if it has non-deleted child attributes when an asset category is deleted all its non-deleted child attributes should be set to `True`(Just like CASCADE on delete). I believe a more efficient way is to make the relationship between `asset category` and `attributes` `many to many`, this way attributes can be sharable among asset categories

api/models/model_operations.py
GbengaOshinaga: On it

GbengaOshinaga:
@SEUNAGBEYE Done

api/models/asset_category.py
ibutiti: Cool @GbengaOshinaga @purpose50 @SEUNAGBEYE

brayoh: Remove the comma at the end

GbengaOshinaga: @brayoh According to the implementation, this function must return a tuple

brayoh: Cool

PR #41, opened 2018-07-09
#158721928 Convert Response object to camel case

frankip:
#### What does this PR do?
Converts all keys in the response object from `snake_case` to `camelCase` 

Currently, the backend makes use of snake casing, while the frontend makes use of camel casing, so the code should be refactored to return camel case to the front-end while also converting the camel case received from the frontend to snake case.

- Fetch the branch `git fetch origin ch-convert-response-to-camel-case-158721928`
- Checkout to the new branch `git checkout ch-convert-response-to-camel-case-158721928`

- Run the application `Flask run`

- Go to any endpoint and check if the 'keys` of the response object is in `camelCase``

We are using Marshmallow's `load_from` and `dump_to` functionalities

[158721928](https://www.pivotaltracker.com/story/show/158721928)

#### Screenshots
**Before**
screen shot 2018-07-05 at 15 07 22

**After**
screen shot 2018-07-06 at 21 05 54


GbengaOshinaga:
@frankip Could you alter the implementation of the GET request to the `api/v1/asset-categories/stats` in the `api.views.asset_category.py` file to make use of marshmallow because right now, sending a GET request to the endpoint still returns `snake_case`

Dheavyman:
@frankip The tests are failing, can you take a look at that to know why.

ibutiti:
You may also need to change the documentation of all endpoints.

api/utilities/model_serializers/asset_category.py
```diff
@@ -36,3 +40,11 @@ class UpdateAssetCategorySchema(BaseSchema):
     """Update Asset category model schema"""
 
     name = fields.String(validate=(string_length_60_validator, name_validator))
+
+class AssetCategoryStatsSchema(BaseSchema):
```
GbengaOshinaga: @frankip I don't think it is necessary for you to create a new schema, you can use the existing `AssetCategorySchema` since it provides the same functionality right now as this one

api/views/asset_category.py
```diff
@@ -78,16 +79,13 @@ def get(self):
...
         asset_categories = AssetCategory._query(request.args)
-        data = []
-        for asset_category in asset_categories:
-            data.append({
-                'id': asset_category.id,
-                'name': asset_category.name,
-                'asset_count': asset_category.assets_count
-            })
+
+        asset_category_schema = AssetCategoryStatsSchema(
```
GbengaOshinaga: @frankip Use the existing `AssetCategorySchema` here

api/utilities/model_serializers/asset_category.py
frankip: Done that

brayoh:
@frankip kinldy squash your commits so that we merge this

PR #45, opened 2018-07-11
#158908009 Eager Load Attributes

GbengaOshinaga:


tests/test_asset_categories_endpoints.py
```diff
@@ -371,3 +371,22 @@ def test_asset_categories_list_endpoint_args(self, client, auth_header):
         asset_category.delete()
         asset_category2.delete()
 
+    def test_eager_load_attributes(self, client, init_db, auth_header, test_asset_category):
```
brayoh: we need a test to check if I send the wrong param I get 0 results back

GbengaOshinaga: @brayoh I checked for that in the method

tests/test_asset_categories_endpoints.py
```diff
@@ -371,3 +371,22 @@ def test_asset_categories_list_endpoint_args(self, client, auth_header):
...
+        response = client.get(f'{api_v1_base_url}/asset-categories?include=attributes',
+                              headers=auth_header)
+        response_json = json.loads(response.data.decode(CHARSET))
+
+        assert response.status_code == 200
+        assert response_json['status'] == 'success'
+        assert isinstance(response_json['data'][0]['customAttributes'], list)
+        assert len(response_json['data'][0]['customAttributes']) is not 0
+
+        response = client.get(f'{api_v1_base_url}/asset-categories?include=attr@@@ibutes',
```
Dheavyman: @GbengaOshinaga I suggest we have this as a separate test case since it's testing for another edge case. Also, the test case name should tell us better what the test wants to check for.

GbengaOshinaga: @brayoh @Dheavyman Done

GbengaOshinaga:
@brayoh @Dheavyman Done

brayoh:
@GbengaOshinaga I've noticed in your postman docs that the response payload has underscores, kindly take care of that and make sure it's camel case

GbengaOshinaga:
@brayoh I actually took care of that, but i think i need to update the link

GbengaOshinaga:
@brayoh I've updated the link, but I have a little issue, because the same endpoint returns a 200 response, I could not get it to return the eager loaded data with the `x-mock-response-code: 200` directive, so for the purpose of it to work, I had to set the response code of the example to `202`, so adding the header `x-mock-response-code: 202` returns the correct data

Pipfile.lock
```diff
@@ -287,7 +287,11 @@
                 "sha256:d345c8fe681115900d6da8d048ba67c25df42973bda370783cd58826442dcd7c",
                 "sha256:e160a7fcf25762bb60efc7e171d4497ff1d8d2d75a3d0df7a21b76821ecbf5c5"
             ],
+<<<<<<< HEAD
             "markers": "python_version >= '2.7' and python_version != '3.2.*' and python_version != '3.1.*' and python_version != '3.0.*' and python_version != '3.3.*'",
+=======
+            "markers": "python_version != '3.1.*' and python_version >= '2.7' and python_version != '3.3.*' and python_version != '3.0.*' and python_version != '3.2.*'",
```
frankip: looks like there are some conflict here

GbengaOshinaga: Thanks @frankip

PR #51, opened 2018-07-17
#159099281 Pagination for assets endpoint

SEUNAGBEYE:
**GIVEN** I have a valid authentication token
**WHEN** I GET to the `/assets?page=<page_number>&limit=<results_limit>` endpoint
**THEN** I will receive paginated results

screen shot 2018-07-17 at 10 56 32 pm
screen shot 2018-07-17 at 10 56 40 pm


api/views/asset.py
```diff
@@ -60,6 +61,23 @@ def post(self):
...
+    @paginate(Asset, AssetSchema)
+    def get(self):
+        """Get Lists of all assets
+        
+        Returns:
+            [JSON] -- [Response object]
+        """
+
+        assets = Asset.query.filter_by(deleted=False).all()
+        assets_data = AssetSchema(many=True).dump(assets)
```
GbengaOshinaga: @SEUNAGBEYE You should exclude fields that are not needed such as `deleted`

api/utilities/paginator.py
```diff
@@ -0,0 +1,108 @@
...
+                    "currentPage": current_page,
+                    "nextPage": next_page,
+                    "previousPage": previous_page,
+                    "pagesCount": page_counts,
+                    "totalCount": records_count
+                }
+
+                return jsonify({
+                    "message": f'{re.sub(r"(?<=[a-z])[A-Z]+",lambda x: f" {x.group(0).lower()}" , model.__name__)} fetched successfully',
+                    "data": schema(many=True).dump(records).data,
```
GbengaOshinaga: Exclude `deleted` here also

GbengaOshinaga:
@SEUNAGBEYE There should not be a link to `previousPage` if the previous page doesn't exist, for example, sending a request to this previous link `localhost:5000/api/v1/asset-categories?page=0&limit=3` gives this error:
```
{
    "message": "Invalid filter column: page",
    "status": "error"
}
```

api/views/asset_category.py
```diff
@@ -235,15 +234,27 @@ def get(self, id):
...
-            exclude=[
-                'deleted_at'
-            ]
-        ).dump(asset_category.assets.filter_by(deleted=False)).data
 
-        return {
-            'status': 'success',
-            'message': serialization_errors['asset_category_assets'].format(asset_category.name),  # noqa
-            'data': assets
+        data, pagination_object = pagination_helper(
```
SEUNAGBEYE: This changes here is not part of the scope of this PR, I just added it to demonstrate how helpful this function is.
I will revert this particular change when the PR is approved.

api/views/asset_category.py
```diff
@@ -196,17 +195,20 @@ class AssetCategoryList(Resource):
 
     @token_required
     @validate_json_request
+    @paginate(AssetCategory, AssetCategorySchema)
```
SEUNAGBEYE: This changes here is not part of the scope of this PR, I just added it to demonstrate how helpful this function is.
I will revert this particular change when the PR is approved.

api/utilities/paginator.py
```diff
@@ -0,0 +1,140 @@
+from functools import wraps
```
Victoradukwu: Please provide a module docstring for this module

api/utilities/paginator.py
```diff
@@ -0,0 +1,140 @@
...
+    else:
+        raise ValidationError({
+            'message':
+            f'{arg_name} contains invalid parameter {arg}'
+        })
+
+
+def pagination_helper(model, schema, extra_query=None):
+
+    limit = validate_args(request.args.get('limit', 'None'), 'limit')
```
Victoradukwu: `request.args.get()` returns `None` by default if `limit` is absent. It's superfluous adding `None` in the function call.

api/utilities/paginator.py
SEUNAGBEYE: @Victoradukwu I will address this.

api/utilities/paginator.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Raises:
+        ValidationError -- [Use to raise exception if any error occur]
+
+    Returns:
+        [bool/int] -- [description]
+    """
+    if arg == 'None':
+        return None
+    if arg.isdigit() and int(arg):
```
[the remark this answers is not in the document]
SEUNAGBEYE: @Victoradukwu valid point made, using `isdigit()` will not crash the app but it will make the program yield incorrect results.

Context: anything `<= 0` is a valid digit, but we don't want to accept data like this `limit=-1&page=0` in the query strings.

Solution: `isdigit()` validates if the `querystring` is a digit, appending `int()` to the condition converts digits `<= 0` to `False` and converts digits `> 0` to `True`.

GbengaOshinaga:
@SEUNAGBEYE The pagination logic doesn't account for additional parameters, for example, sending a request to the endpoint: `localhost:5000/api/v1/asset-categories?page=1&limit=5&include=attributes` paginates the data, but doesn't return the required attributes. I believe this is not the expected behavior. Any thoughts?

api/utilities/paginator.py
```diff
@@ -0,0 +1,164 @@
...
+                
+            Returns:
+                [type] -- [description]
+            """
+
+            data, pagination_object = pagination_helper(model, schema)
+
+            if data and pagination_object:
+
+                return jsonify({
```
frankip: Can the marshmallow validator `dump` be used here instead of the jsonify?

SEUNAGBEYE: @frankip  Done

SEUNAGBEYE:
@GbengaOshinaga Your suggestion could not be implemented at the moment because the `DynamicFilter` currently has `BUG`.

api/utilities/paginator.py
```diff
@@ -0,0 +1,208 @@
+"""Module for pagination helpers"""
+
+from functools import wraps
+from math import ceil
+import re
+
+from flask import request
+
+from .messages.error_messages import serialization_errors
+
```
Dheavyman: This line space is not necessary since they are both local application modules

SEUNAGBEYE: @Dheavyman `request` is a third party import and  `serialization_errors` is a local `application/library` specific import.

According to PEP8 "Imports should be grouped in the following order":
```
Standard library imports.
Related third party imports.
Local application/library specific imports.
```
According to StackOverflow the `local` in the third statement "applies to both application and library imports of your own (eg. your own libraries/classes)".

Please check this https://stackoverflow.com/questions/16759244/pep-8-library-specific-imports for more clarity because there is a lot of ambiguity in the third statement

Dheavyman: @SEUNAGBEYE I am referring to the space on line 10. It is between two local application modules which should be grouped together. I understand what they mean by the levels of import. There should be no line space on line 10.

SEUNAGBEYE: @Dheavyman Fixed. I thought you were talking about line `7 & 9`.

api/utilities/paginator.py
```diff
@@ -0,0 +1,206 @@
...
+from math import ceil
+import re
+
+from flask import request
+
+from .messages.error_messages import serialization_errors
+from ..middlewares.base_validator import ValidationError
+
+
+def validate_args(arg, arg_name):
```
Dheavyman: @SEUNAGBEYE Nice. If this function is just to validate args that have to deal with pagination, can we have it named something like `validate_pagination_args`? I believe this will make it more distinct, descriptive and easier to read and understand what the function intends to achieve.

api/utilities/paginator.py
```diff
@@ -0,0 +1,206 @@
...
+    Returns:
+        (bool/int) -- Returns True or False
+    """
+
+    # Sets limit to a default of 10 if it is not part of the query string
+    if arg_name == 'limit' and arg == 'None':
+        limit = 10
+        return limit
+
+    # Returns false of query strings are missing
```
Dheavyman: Typo in the comment, I believe you meant `if` and not `of`

api/utilities/paginator.py
```diff
@@ -0,0 +1,206 @@
...
+from .messages.error_messages import serialization_errors
+from ..middlewares.base_validator import ValidationError
+
+
+def validate_args(arg, arg_name):
+    """
+    Validates if the query strings are valid.
+
+    Arguments:
+        arg (string): Query string
```
Dheavyman: There are two arguments in the function but only one is documented here. Please review

SEUNAGBEYE: @Dheavyman Done.

api/utilities/paginator.py
SEUNAGBEYE: @Dheavyman Done

api/utilities/paginator.py
SEUNAGBEYE: @Dheavyman Done

api/utilities/paginator.py
brayoh: @SEUNAGBEYE ☝️

SEUNAGBEYE: @brayoh This had already been addressed before you left this message

SEUNAGBEYE:
@brayoh Feedback implemented.

PR #52, opened 2018-07-17
#159102652 Remove Content-Type Validation From GET Requests

GbengaOshinaga:
```
{
    "message": "Content-Type should be application/json",
    "status": "error"
}
```
which is not ideal for GET requests. This PR fixes this problem.


ibutiti:
Your PR description is very sketchy. Keep in mind that this PR will be read by someone many months down the line who doesn't know anything about the background of this task and will rely on this info to get context.

GbengaOshinaga:
@ibutiti Added more context to the PR description

PR #54, opened 2018-07-18
 #158660348 Add bake image and deploy scripts

anyric:
#### What does this PR do?
This PR adds bake image and deployment scripts in the CircleCi CI/CD workflows for Activo Api hence deploying it to GCP infrastructure.

- Add bake image script to prepare deployment image
- Add integration and deployment script for Activo Api

This is handled automatically by CircleCi once a commit is pushed to the repo

Activo Api initially do not have bake image and deployment scripts in its CircleCi workflow hence this PR integrates the bake image and deployment scripts in the CircleCi workflow for deploying Activo Api to GCP infrastructure.

[#158660348](https://www.pivotaltracker.com/story/show/158660348)
[#158660723](https://www.pivotaltracker.com/story/show/158660723)
N/A

#### Questions:
N/A


.circleci/bake.sh
```diff
@@ -0,0 +1,58 @@
...
+    ENV_NAME="$DEPLOYMENT_ENVIRONMENT" HOME_PATH="/home/circleci/activo-api" PROJECT_ID="$GCLOUD_ACTIVO_PROJECT" packer build packer.json 2>&1 | tee packer_output.log
+    PACKER_IMG_TAG="$(grep 'A disk image was created:' packer_output.log | cut -d' ' -f8)"
+  popd
+  mkdir -p workspace
+  echo $PACKER_IMG_TAG > ~/activo-api/workspace/output
+  cat ~/activo-api/workspace/output
+}
+
+check_out_infrastructure_code() {
+  echo "Checkout the infrastructure code"
```
brayoh: I think we should have as `Checking the infrastructure code`

.circleci/deploy.sh
```diff
@@ -0,0 +1,119 @@
...
+  EMOJI=${EMOJIS[$RANDOM % ${#EMOJIS[@]} ]}
+  COMMIT_LINK="https://github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/commit/${CIRCLE_SHA1}"
+  DEPLOYMENT_TEXT="Tag: ${PACKER_IMG_TAG} has just been deployed as the latest ${PROJECT} in ${DEPLOYMENT_ENVIRONMENT}  $COMMIT_LINK "
+  DEPLOYMENT_CHANNEL=${SLACK_CHANNEL}
+  IMG_TAG="$(git rev-parse --short HEAD)"
+  SLACK_DEPLOYMENT_TEXT="Git Commit Tag: <$COMMIT_LINK|${IMG_TAG}> has just been deployed to *${PROJECT}* in *${DEPLOYMENT_ENVIRONMENT}* ${EMOJI}"
+}
+
+check_out_infrastructure_code() {
+  echo "Checkout infrastructure code"
```
brayoh: same here

config.py
```diff
@@ -18,6 +18,10 @@ class Config(object):
     DEBUG = False
     TESTING = False
 
+class StagingConfig(Config):
+    """App staging configuration."""
+
+    pass
```
brayoh: since staging and development share most parts of the config, let's have that here instead of having pass

.circleci/bake.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+set -ex
+set -o pipefail
+
+echo "building the packer image"
+declare_env_variables() {
```
brayoh: Add a new line here before the function declaration

.circleci/deploy.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+
+set -ex
+set -o pipefail
+
+echo "Declaring environment variables"
+declare_env_variables() {
```
brayoh: New line here too

anyric:
i did a rebase on develop and during fixing merge conflict as well as combining commit i squashed commit message to the merge message so his commit happens to be part of the merge commit that is why he is appearing there.

brayoh:
His commit shouldn't be part of your commit, just `git reflog` then choose the `HEAD` before you squashed and `git reset --soft HEAD@{<number>}` then add your changes, commit and squash again.

config.py
anyric: i think we can even remove that StagingConfig class and map staging in the config dictionary to ProductionConfig instead of development since both inherits from Config class only

brayoh: Does staging share the same config with production?

brayoh: Also is the `JWT_SECRET_KEY` included in the staging config?

anyric: yes staging share the same config with production and it includes the JWT_PUBLIC_KEY which is passed as environment variable through CircleCi online settings

anyric:
Suggested changes implemented.

brayoh:
@anyric let's get more reviews on this

config.py
```diff
@@ -65,6 +65,7 @@ class TestingConfig(Config):
 
 config = {
   'development': DevelopmentConfig,
+  'staging': ProductionConfig,
```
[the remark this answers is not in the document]
anyric: Initially there was a StagingConfig class mapped to staging but just having pass since there were no new declarations in it apart from the inheritance from Config class, so there was a suggestion to use development Config instead of using pass. But since staging is to mimic production, then its better to have them share the same config with production than with development which has debugging mode enabled.

.circleci/config.yml
```diff
@@ -121,3 +212,69 @@ workflows:
...
+      - deploy_sandbox:
+          requires:
+            - release_to_sandbox
+          filters:
+            branches:
+              only: /^(.*?(\bsandbox\b)[^$]*)$/
+      - release_to_staging:
+          requires:
+            - run_tests
+            - upload_coverage
```
indungu: Having both the `run_tests` and the `upload_coverage` jobs as requisite jobs for this workflow seems a bit redundant since the latter already requires the former

.circleci/config.yml
```diff
@@ -121,3 +212,69 @@ workflows:
...
+          requires:
+            - release_to_staging
+          filters:
+            branches:
+              only: develop
+            tags:
+              only:
+                - /v[0-9]+(\.[0-9]+)*/
+      - release_to_production:
+          requires:
```
indungu: Should't this job require the `upload_coverage` job instead of the `run_tests` as sited in the review of lines 244 and 245?

anyric: suggestions implemented

config.py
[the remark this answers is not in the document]
Thegaijin: While staging and production have the same configurations now, what if you want to test out a change in the configurations on staging before adding it to production?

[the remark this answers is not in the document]
anyric: The environments are customisable through environment variables so it will not block us from making changes to both staging and production

Thegaijin:
LGTM

PR #62, opened 2018-07-23
#159192785 Create centers

GbengaOshinaga:
```
{
    "name": "Nairobi",
    "image": {
        "public_id": "cr4mxeqx5zb8rlakpfkg",
        "version": 1372275963,
        "signature": "63bfbca643baa9c86b7d2921d776628ac83a1b6e",
        "width": 864,
        "height": 576,
        "format": "jpg",
        "resource_type": "image",
        "created_at": "2017-06-26T19:46:03Z",
        "bytes": 120253,
        "type": "upload",
        "url": "https://res.cloudinary.com/demo/image/upload/v1372275963/cr4mxeqx5zb8rlakpfkg.jpg",
        "secure_url": "https://res.cloudinary.com/demo/image/upload/v1372275963/cr4mxeqx5zb8rlakpfkg.jpg"
    }
}
```
>- You should receive a success response with the following payload:
```
{
    "status": "success",
    "message": "Center created successfully",
    "data": {
        "name": "Nairobi",
        "id": "-LI6SDidpb7RS-1b4Tx8",
        "image": {
            "public_id": "cr4mxeqx5zb8rlakpfkg",
            "version": 1372275963,
            "signature": "63bfbca643baa9c86b7d2921d776628ac83a1b6e",
            "width": 864,
            "height": 576,
            "format": "jpg",
            "resource_type": "image",
            "created_at": "2017-06-26T19:46:03Z",
            "bytes": 120253,
            "type": "upload",
            "url": "https://res.cloudinary.com/demo/image/upload/v1372275963/cr4mxeqx5zb8rlakpfkg.jpg",
            "secure_url": "https://res.cloudinary.com/demo/image/upload/v1372275963/cr4mxeqx5zb8rlakpfkg.jpg"
        }
    }
}
```


api/views/center.py
```diff
@@ -0,0 +1,39 @@
...
+    @token_required
+    @validate_json_request
+    def post(self):
+        """
+        POST method for creating centers.
+
+        Payload should have the following parameters:
+            name(str): name of the center
+            image(dict): image meta data
+        """
```
brayoh: Nice!

api/views/center.py
```diff
@@ -0,0 +1,39 @@
...
+
+        request_data = request.get_json()
+        center_schema = CenterSchema(exclude=['deleted'])
+        center = center_schema.load_object_into_schema(request_data)
+        center.save()
+
+        return {
+            'status': 'success',
+            'message': SUCCESS_MESSAGES['created'].format('Center'),
```
brayoh: good stuff

tests/mocks/center.py
```diff
@@ -0,0 +1,61 @@
...
+        "2017-06-26T19:46:03Z",
+        "bytes":
+        120253,
+        "type":
+        "upload",
+        "url":
+        "https://res.cloudinary.com/demo/image/upload/v1372275963/cr4mxeqx5zb8rlakpfkg.jpg",
+        "secure_url":
+        "https://res.cloudinary.com/demo/image/upload/v1372275963/cr4mxeqx5zb8rlakpfkg.jpg"
+    }
```
brayoh: Fix the formatting for some of the keys, they shouldn't be on a new line.

tests/mocks/center.py
```diff
@@ -0,0 +1,61 @@
...
```
brayoh: same here

tests/test_center_endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_create_center_failure_with_invalid_data(self, client,
+                                                     auth_header):
+        """
+        Test that an error is returned when creating
+        a center with invalid data
+
+        Parameters:
+            client: fixture to get flask test client
+            auth_header: fixture to get token
+        """
```
brayoh: Add the data type for the parameters

tests/test_center_endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_create_center_failure_with_incomplete_data(
+            self, client, auth_header):
+        """
+        Test that an error is returned when creating
+        a center with incomplete data
+
+        Parameters:
+            client: fixture to get flask test client
+            auth_header: fixture to get token
+        """
```
brayoh: Add the data types here too

tests/test_center_endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+        assert response_json['status'] == 'error'
+        assert response_json['message'] == 'The center already exists'
+
+    def test_create_center_failure_without_token(self, client):
+        """
+        Test that an error is returned when trying to
+        create a center without a valid token
+
+        Parameters:
+            client: fixture to get flask test client
```
brayoh: Same here

tests/test_center_endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_create_center_failure_with_existing_data(self, client,
+                                                      auth_header):
+        """
+        Test that error is returned when
+        creating an already existing center
+
+        Parameters:
+            client: fixture to get flask test client
+            auth_header: fixture to get token
+        """
```
brayoh: And here

tests/test_center_endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def test_create_center_success(self, client, auth_header, init_db):
+        """
+        Test successfully creating a center
+
+        Parameters:
+            client: fixture to get flask test client
+            auth_header: fixture to get token
+            init_db: fixture to initialize the test database
+        """
```
brayoh: Same here

tests/mocks/center.py
GbengaOshinaga: @brayoh I agree, but the pre-commit hook is responsible for this formatting

tests/mocks/center.py
GbengaOshinaga: @brayoh I don't understand

api/views/center.py
GbengaOshinaga: @brayoh Thanks

tests/mocks/center.py
brayoh: Cool

tests/mocks/center.py
brayoh: Sorry, ignore the above comment, I meant fix indentation

brayoh:
@GbengaOshinaga get more reviews for this

api/utilities/model_serializers/center.py
```diff
@@ -0,0 +1,29 @@
...
+        error_messages={'required': serialization_errors['field_required']})
+    image = fields.Dict(required=True)
+
+    @post_load
+    def create_center(self, data):
+        """Return center object after successful loading of data"""
+        result = Center.query.filter_by(name=data['name']).first()
+        if not result:
+            return Center(**data)
+        raise ValidationError({'message': 'The center already exists'}, 409)
```
Dheavyman: @GbengaOshinaga I think having a message like `Center with the name <name> already exists` communicates the reason for the error better.

GbengaOshinaga: @Dheavyman Done

api/models/center.py
```diff
@@ -0,0 +1,29 @@
+from sqlalchemy.dialects.postgresql import JSON
+
+from .base.base_model import BaseModel
+from .database import db
+
+
+class Center(BaseModel):
```
Victoradukwu: @GbengaOshinaga I believe this should inherit from AuditableModel so we can have other relevant fields. Consider this for the  `center` serializer too.

GbengaOshinaga: @Victoradukwu I'm not so certain this model needs the fields provided by the AuditableBaseModel, however, @brayoh can you please clarify?

ibutiti: I agree

api/models/center.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Model for centers
+    """
+
+    __tablename__ = 'centers'
+
+    name = db.Column(db.String(60), nullable=False)
+    image = db.Column(JSON, nullable=False)
+    assets = db.relationship('Asset', backref='center', cascade='delete')
```
ibutiti: I think we agreed to remove the cascade options so the application throws an error when anyone tries to delete it.

api/models/center.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Center(BaseModel):
+    """
+    Model for centers
+    """
+
+    __tablename__ = 'centers'
+
+    name = db.Column(db.String(60), nullable=False)
```
ibutiti: Because our centres are/will be so few and with very few changes, should we make this unique?

api/utilities/model_serializers/center.py
```diff
@@ -0,0 +1,33 @@
...
+
+from api.models.center import Center
+from api.middlewares.base_validator import ValidationError
+from .base_schemas import BaseSchema
+from ..validators.name_validator import name_validator
+from ..messages.error_messages import serialization_errors
+from ..validators.string_length_validators import string_length_60_validator
+
+
+class CenterSchema(BaseSchema):
```
ibutiti: See victor's comment on `auditable base schema`

api/utilities/model_serializers/center.py
```diff
@@ -0,0 +1,33 @@
...
+        required=True,
+        validate=(string_length_60_validator, name_validator),
+        error_messages={'required': serialization_errors['field_required']})
+    image = fields.Dict(required=True)
+
+    @post_load
+    def create_center(self, data):
+        """Return center object after successful loading of data"""
+        name = data['name']
+        result = Center.query.filter_by(name=name).first()
```
ibutiti: I'm of the opinion that this validation should happen at the endpoint stage because we should be able to use this schema for the PATCH endpoint as well. 

I think this post load method should only change the name to Title case and not return the instance because we won't need a new instance in the case of a PATCH

api/views/center.py
ibutiti: Nice

api/utilities/model_serializers/center.py
ibutiti: We should also include the deleted flag in the filter.

api/views/center.py
ibutiti: It would be good to add in the name of the centre in this message text.

api/models/asset.py
```diff
@@ -10,9 +10,9 @@ class Asset(AuditableBaseModel):
     tag = db.Column(db.String(60), nullable=False, unique=True)
     serial = db.Column(db.String(60), nullable=True)
     custom_attributes = db.Column(JSON, nullable=True)
-    asset_category_id = db.Column(db.String,
-                                  db.ForeignKey('asset_categories.id'),
-                                  nullable=False)
+    asset_category_id = db.Column(
+        db.String, db.ForeignKey('asset_categories.id'), nullable=False)
+    center_id = db.Column(db.String, db.ForeignKey('centers.id'))
```
ibutiti: Is there a reason this is nullable?

api/models/user.py
```diff
@@ -12,6 +12,7 @@ class User(BaseModel):
     name = db.Column(db.String(60), nullable=False)
     email = db.Column(db.String(60), nullable=False, unique=True)
     image_url = db.Column(db.String)
+    center_id = db.Column(db.String, db.ForeignKey('centers.id'))
```
ibutiti: Is there a reason this is nullable?

tests/test_center_endpoints.py
```diff
@@ -0,0 +1,114 @@
...
+
+        center = json.dumps(VALID_CENTER)
+        response = client.post(
+            f'{API_V1_BASE_URL}/centers', data=center, headers=auth_header)
+        response_json = json.loads(response.data.decode(CHARSET))
+        assert response.status_code == 409
+        assert response_json['status'] == 'error'
+        assert response_json[
+            'message'] == 'Center with the name Lagos already exists'
```
ibutiti: Please link this to the error message object so we don't have to change tests when we alter messages.

tests/test_center_endpoints.py
```diff
@@ -0,0 +1,114 @@
...
+            client(object): fixture to get flask test client
+        """
+
+        center = json.dumps(VALID_CENTER)
+        response = client.post(f'{API_V1_BASE_URL}/centers', data=center)
+        response_json = json.loads(response.data.decode(CHARSET))
+        assert response.status_code == 401
+        assert response_json['status'] == 'error'
+        assert response_json[
+            'message'] == 'Bad request. Header does not contain an authorization token.'
```
ibutiti: See above messages comment.

tests/test_center_model.py
```diff
@@ -0,0 +1,28 @@
+"""Module for center model tests"""
+
+from api.models import Center
+
+
+class TestCenterModel:
+    def test_new_center(self, new_center, init_db):
+        assert new_center == new_center.save()
```
ibutiti: Please query the db for the count of centers to verify it's 1 as expected.

tests/test_center_model.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def test_update(self, new_center):
+        new_center._update(name='Nairobi')
+        assert new_center.name == 'Nairobi'
```
ibutiti: Please make a query on this one using the id generated by `new_center` to ascertain that we're not checking for the object in the session and that it's been committed to the db.

tests/test_center_model.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def test_get(self, new_center):
+        assert Center.get(new_center.id) == new_center
+
+    def test_count(self, new_center):
```
ibutiti: this can be combined with the first test above.

tests/test_center_model.py
```diff
@@ -0,0 +1,28 @@
...
+        assert new_center.count() == 1
+
+    def test_query(self, new_center):
+        center_query = new_center._query()
+        assert center_query.count() == 1
+        assert isinstance(center_query.all(), list)
+
+    def test_delete(self, new_center, request_ctx,
+                    mock_request_obj_decoded_token):
+        new_center.delete()
```
ibutiti: your `new_center` from the fixtures has not been saved. This doesn't test the delete unless your tests have side effects.

api/models/center.py
Dheavyman: @GbengaOshinaga I think you need to seek for clarification on this. @ibutiti He added the delete on cascade because of the what the design on invision stipulated. The design warns that deleting a center will delete the corresponding users and assets. But from my sync with Victor on his task which involves deleting a center, a center cannot be deleted if it still has users or assets associated with it.

Dheavyman: Also, very important is adding the lazy parameter to the relationships, we need it to return a query object and not data object so as to be able to perform further query on it, like filter out the deleted ones or so. This also reduces making heavy database transactions.

api/models/center.py
```diff
@@ -0,0 +1,29 @@
...
+    image = db.Column(JSON, nullable=False)
+    assets = db.relationship('Asset', backref='center', cascade='delete')
+    users = db.relationship('User', backref='center', cascade='delete')
```
Dheavyman: The lazy argument is needed here too. Same reason as above. To add to that, the delete validator requires you pass a query object to it and not the actual data.

api/models/asset.py
GbengaOshinaga: @ibutiti If nullable is set to false, then it would be impossible to add assets without adding the center id, so it is nullable to allow the frontend the ability to add assets without specifying the center id, since they have not implemented that functionality yet

api/models/center.py
ibutiti: Yeah, that design different from how we've been doing things.... you should seek clarification @GbengaOshinaga

api/models/asset.py
ibutiti: Hmmm... okay. Please add a chore to fix this so it doesn't fall through the cracks.

api/models/center.py
ibutiti: I agree.

api/models/asset.py
GbengaOshinaga: Alright

api/views/center.py
GbengaOshinaga: @ibutiti Since the name of the center is visibly present in the data returned, I don't think it is necessary to add it

GbengaOshinaga:
All comments have been addressed

api/views/center.py
ibutiti: That's true, however, i think the front end is directly passing the message to the user so this may be useful.

GbengaOshinaga: @ibutiti I discussed it with the front-end devs, and they agreed that this message is most appropriate for them.

SEUNAGBEYE: @ibutiti I think Center created is good because it is short and clear.

tests/mocks/center.py
SEUNAGBEYE: We need to do something to this pre-commit hook, probably disable this kind of formatting.

PR #65, opened 2018-07-24
#159232693 Modify response data for choices

kenoseni:
#### What does this PR do?
Ensures choices are sent as an array of string to the client
When a post request is made to `/asset-categories` the choices attribute, if available, should be an array of string
- clone the repo `git pull origin ch-change-choices-to-array-159232693`
- checkout to the branch 'git checkout ch-change-choices-to-array-159232693`
- make a post request to the asset categories
`{{base-url}}/api/v1/asset-categories`
- though the choices attributes are stored as a string in the database, they are now returned as an array of string to the client
[#159232693](https://www.pivotaltracker.com/story/show/159232693)
#### Postman Collection
[Link](https://www.getpostman.com/collections/0ad0ae70f5a4bd8a80af)


api/utilities/model_serializers/attribute.py
```diff
@@ -59,33 +52,57 @@ def create_attribute(self, data):
...
+        Return attribute object after successful loading into schema
+        """
+
+        choices = data['choices']
+        if not choices or choices == ['']:
+            choices = []
+        else:
+            choices = choices[0].split(',')
+
+        data['choices'] = choices
```
ibutiti: Why the `choices` variable? You can directly mutate the `data['choices']` object.

api/utilities/model_serializers/attribute.py
```diff
@@ -59,33 +52,57 @@ def create_attribute(self, data):
 
         return Attribute(**data)
 
+    @post_dump
+    def dump_attribute(self, data):
+        """
+        Return attribute object after successful loading into schema
```
ibutiti: Please elaborate further, the purpose of this function is not clear.

api/utilities/model_serializers/attribute.py
```diff
@@ -10,42 +11,34 @@
...
+    input_control = fields.String(
+        required=True,
+        load_from='inputControl',
+        dump_to='inputControl',
+        validate=(string_length_60_validator, name_validator,
+                  input_control_validation),
+        error_messages={
+            'required': serialization_errors['field_required']  # noqa
+        })
+    choices = fields.List(fields.String(validate=(string_length_60_validator)))
```
ibutiti: This validation is inconsistent with the update schema. Is there a reason you changed it from the 250 length here?

api/utilities/validators/attributes_validator.py
```diff
@@ -31,11 +31,34 @@ def validate_attribute_exists(attribute_id):
             attribute_id=attribute_id, asset_category_id=asset_category_id))
 
 
+def remove_duplicate(choices):
+    """
+    Remove duplicates from choices
+    """
+
+    unique_choices = []
+    for choice in choices:
```
ibutiti: This can be achieved with the `set()` method. Also, you may need to control for different cases. Dropdown and dropdown are different in Python.

api/utilities/model_serializers/attribute.py
Dheavyman: @kenoseni The update-schema should be modified to this instead. The update-schema validation was for when the choice was a string field and not an array. This validation here is for each item of the array.

api/utilities/validators/attributes_validator.py
SEUNAGBEYE: @ibutiti the `set()` method would re-arrange the array, and some tests would be failing if `set()` is used.

Context:
`[1,2,3] != [2,1,3]`

api/utilities/validators/attributes_validator.py
```diff
@@ -31,11 +31,34 @@ def validate_attribute_exists(attribute_id):
...
+        if choice and choice not in unique_choices:
```
SEUNAGBEYE: @kenoseni I suggest you do this: `if choice and choice.lower() not in unique_choices:` because `Red != red`

api/utilities/validators/attributes_validator.py
ibutiti: Agreed @SEUNAGBEYE. Checkout the [Counter()](https://stackoverflow.com/questions/7828867/how-to-efficiently-compare-two-unordered-lists-not-sets-in-python) method

andela-kakpobome:
1. The Pivotal tracker ID should be a link to the actual story, I've updated your PR, but ensure you do that going forward

2. There's still no link to the updated documentation, why?

purpose50:
@kenoseni your test is failing 👆

api/utilities/model_serializers/attribute.py
brayoh: @kenoseni ☝️

api/utilities/validators/attributes_validator.py
brayoh: @kenoseni ☝️

kenoseni: The Set Method does not solve the problem at hand because  It does not take care of choice values which are differentiated by upper or lower cases.
Tests fail if we make use of the set method and using the counter method in the tests does not solve the problem of duplicate choice values with different cases.
Context:
['Red', 'red', 'REd'] cannot be handled by set and counter method even though the counter method ensures that test assertions pass.

api/utilities/model_serializers/attribute.py
kenoseni: @brayoh This has already been taken care of. The choices field was changed from string to list and do not require the initial validation of a string(length 250)

brayoh: Cool

api/utilities/validators/attributes_validator.py
brayoh: 👍 makes sense

brayoh:
@kenoseni Your tests are failing, seems to be an issue with `Centers` model

brayoh:
@kenoseni Check the response message under the `400` request example in the `postman collection` for `get a list of asset categories` and update it appropriately.

kenoseni:
@brayoh the response message in the postman collection has been updated to a 401 and all tests are passing. Thanks

PR #67, opened 2018-07-26
#159191866 Update circleci to use pipfile for dependencies

anyric:
#### What does this PR do?
This PR updates project dependency management in requirements.txt to pipfile and rename terraform jwt_public_key variable.

- Update CircleCI configuration to use pipfile instead of requirements.txt file 

This is handled automatically by CircleCi once a commit is pushed to the repo

Currently the project has both requirements.txt and pipfile to manage dependencies. Using pipfile for dependencies management is preferable as it can separate development and production dependencies in an elegant and easy to manage way

[#159191866](https://www.pivotaltracker.com/story/show/159191866)

N/A

#### Questions:
N/A


.circleci/config.yml
```diff
@@ -79,7 +84,7 @@ jobs:
       - run:
           name: Set up database
           command: |
-            . venv/bin/activate
+            source $(python3 -m pipenv --venv)/bin/activate
```
ibutiti: You can execute commands within the pipenv virtualenv by executing  `pipenv run <command>`. Please see [this article](https://robots.thoughtbot.com/how-to-manage-your-python-projects-with-pipenv)

.circleci/config.yml
```diff
@@ -97,7 +102,7 @@ jobs:
       - run:
           name: Run tests
           command: |
-            . venv/bin/activate
+            source $(python3 -m pipenv --venv)/bin/activate
```
ibutiti: `pipenv shell` activates the associated virtualenv. Is there a reason you're using this command?

.circleci/config.yml
anyric: running pipenv shell to activate the virtualenv gives pipenv not found error, the above is an alternative way to activate which works.

.circleci/config.yml
anyric: pipenv shell launches a subshell environment hence app and postgres database can't communicate and fails with timeout error so that alternative command is the one that works since it activates the already created environment in which migrations connection exists

anyric: @ibutiti check this issue for context https://github.com/pypa/pipenv/issues/773

.circleci/deploy.sh
```diff
@@ -75,11 +75,11 @@ build_infrastructure() {
     if [ "$DEPLOYMENT_ENVIRONMENT" == "production" ]; then
       terraform apply --parallelism=1 -var="api_state_path=${TF_VAR_state_path}" -var="api_project_id=${TF_VAR_project}" -var="api_bucket=${TF_VAR_bucket}" -var="api_env_name=${DEPLOYMENT_ENVIRONMENT}" -var="api_flask_env=${FLASK_ENV}" -var="api_disk_image=${PACKER_IMG_TAG}" -var="api_reserved_env_ip=${RESERVED_IP}" \
```
nyambati: Just an observation, can we reduce the `-var`  by use of `.tfvars` files https://www.terraform.io/intro/getting-started/variables.html#from-a-file

.circleci/config.yml
```diff
@@ -37,20 +37,25 @@ cmd_install_dependencies: &cmd_install_dependencies
...
+        key: api-dependencies-{{ checksum "Pipfile" }}
         paths:
-          - ./venv
+          - $(python3 -m pipenv --venv)
 
 cmd_restore_cache: &cmd_restore_cache
     restore_cache:
         keys:
-          - api-dependencies-{{ checksum "requirements.txt" }}
+          - api-dependencies-{{ checksum "Pipfile" }}
```
SEUNAGBEYE: Nice

.circleci/deploy.sh
nyambati: ok

PR #68, opened 2018-07-26
#159192999 Delete a single center

Victoradukwu:


api/utilities/messages/success_messages.py
```diff
@@ -3,5 +3,8 @@
 SUCCESS_MESSAGES = {
     'asset_created': 'Asset successfully created',
     'asset_edited': 'Asset successfully edited',
-    'created': '{} successfully created'
+    'created': '{} successfully created',
+    'deleted': '{0} successfully deleted',
+    'hard_delete': '{} center successfully hard deletd',
```
GbengaOshinaga: @Victoradukwu You misspelled deleted here and the line below

Victoradukwu:
@GbengaOshinaga Thanks for spotting that. I have fixed it.

tests/mocks/center.py
```diff
@@ -31,6 +31,36 @@
     }
 }
 
+VALID_CENTER_FOR_DELETE = {
```
GbengaOshinaga: @Victoradukwu Why not use the already existing `VALID_CENTER` mock, instead of creating another identical one

api/utilities/center_helpers.py
```diff
@@ -0,0 +1,22 @@
+import datetime
+
+from flask import request
+from ..models.center import Center
+from .messages.success_messages import SUCCESS_MESSAGES
+from api.models.database import db
+
+
+def delete_center(id):
```
GbengaOshinaga: @Victoradukwu I am of the opinion that since this method is specific to just deleting a center, the logic should be implemented in the endpoint and not in a separate module in the `utilities` package

tests/test_center_endpoints.py
```diff
@@ -110,3 +113,102 @@ def test_create_center_failure_with_invalid_data(self, client,
...
+            self, client, init_db, auth_header):  #pylint: disable=W0613
+        """
+        Tests that 400 is returned when id is invalid
+        """
+        response = client.delete(
+            f'{API_V1_BASE_URL}/centers/-LX%%%tghrfe5', headers=auth_header)
+        response_json = json.loads(response.data.decode(CHARSET))
+        assert response.status_code == 400
+        assert response_json['status'] == 'error'
+        assert response_json['message'] == 'Invalid id in parameter'
```
GbengaOshinaga: @Victoradukwu Use `serialization_errors['invalid_id']` here instead of hard-coding the message since it is already present in the `error_messages` module

tests/test_center_endpoints.py
```diff
@@ -110,3 +113,102 @@ def test_create_center_failure_with_invalid_data(self, client,
...
+            self, client, init_db):  #pylint: disable=W0613
+        """
+        Tests that 401 is returned when token is not provided
+        """
+        response = client.delete(f'{API_V1_BASE_URL}/centers/-LX%%%tghrfe5')
+        response_json = json.loads(response.data.decode(CHARSET))
+        assert response.status_code == 401
+        assert response_json['status'] == 'error'
+        assert response_json[
+            'message'] == 'Bad request. Header does not contain an authorization token.'
```
GbengaOshinaga: @Victoradukwu Same here, use `jwt_errors['NO_TOKEN_MSG']` instead

api/utilities/center_helpers.py
```diff
@@ -0,0 +1,22 @@
...
+    message = ''
```
GbengaOshinaga: It's redundant to declare and initialize `message` here since it will be declared and initialized in either the `if` or `else` block

tests/mocks/center.py
Victoradukwu: That would fail due to the unique constraint on the center `name`.

api/utilities/center_helpers.py
Victoradukwu: The extraction is to improve readability and avoid cluttering the view with too many lines of code.

tests/test_center_endpoints.py
Victoradukwu: @GbengaOshinaga Thanks for these. I have handled them.

api/utilities/center_helpers.py
GbengaOshinaga: @Victoradukwu I understand your point, but any function created in the `utilities` package should be reusable, but this function is specific to just deleting a center, therefore it should be implemented in the endpoint, and not in `utilities`

purpose50: @Victoradukwu the delete functionality is specific to the Center and it is not a utility, the logic should be in the function for the endpoint, Functions in utilities are meant to be reusable, they are like tools that are used in different parts of the codes. the delete_center function you created can't be reused for any other models and is specific just to the Center model, it is best for the logic to be in the function for the endpoint.

api/utilities/helpers/center_helpers.py
```diff
@@ -0,0 +1,30 @@
+"""Module for deleting a center"""
+import datetime
+
+from flask import request  #pylint:disable=E0401
+from api.models.database import db
+from ..models.center import Center  #pylint:disable=E0401
```
purpose50: @Victoradukwu lets do our best to avoid comments like this: `#pylint:disable=E0401`, it makes the code looks unclean when they are everywhere.

tests/test_center_endpoints.py
```diff
@@ -110,3 +113,101 @@ def test_create_center_failure_with_invalid_data(self, client,
...
+        """
+            Tests that 404 is returned for an attempt to delete a non-existent center
+        """
+
+        response = client.delete(
+            f'{API_V1_BASE_URL}/centers/-L6YY54rfd98GTY', headers=auth_header)
+        response_json = json.loads(response.data.decode(CHARSET))
+        assert response.status_code == 404
+        assert response_json['status'] == 'error'
+        assert response_json['message'] == 'Center not found'
```
SEUNAGBEYE: @Victoradukwu Please create an error message for this in the `error_message.py`, that may require you change the `get_or_404` response to use the error_message that would be created in the `error_messages.py`.

tests/test_center_endpoints.py
```diff
@@ -110,3 +113,103 @@ def test_create_center_failure_with_invalid_data(self, client,
...
+        assert response_json['message'] == SUCCESS_MESSAGES[
+            'hard_delete'].format(test_center.name)
+
+    def test_soft_delete_center_with_user(  #pylint:disable=C0103
+            self, client, init_db, test_center, auth_header):  # pylint: disable=W0613
+        """
+            Tests for soft deleting a center with an associated user
+        """
+        User(
+            id='-GTHFDR56765',
```
SEUNAGBEYE: @Victoradukwu I suggest we have this as a mock data.

Victoradukwu:
@GbengaOshinaga @purpose50 @SEUNAGBEYE All requested changes have been applied.

PR #74, opened 2018-07-30
#159231737 Endpoint to get users under a center

ibutiti:
#### What does this PR do?
- Adds an endpoint to get all non-deleted users in a center.

**GIVEN** I have a valid authentication token
**WHEN** I `GET` to the `/centers/{<center_id>}/people` endpoint
**THEN** I will receive a list of all people under that center

- Create a local copy with the following commands: `git fetch origin ft-get-people-in-center-endpoint-159231737:ft-get-people-in-center-endpoint-159231737`, then `git checkout ft-get-people-in-center-endpoint-159231737`
- Run the new migration scripts by executing `flask db upgrade`
- Activate your virtual environment by executing `pipenv shell`
- Using postman, send a `GET` request to: `localhost:5000/api/v1/centers/<center_id>/people`

- N/A

- [#159231737](https://www.pivotaltracker.com/story/show/159231737)

#### Postman collection:
- [Link](https://www.getpostman.com/collections/497804071cea71ad26f9)

- N/A

#### Questions:
- N/A


api/models/user.py
```diff
@@ -13,6 +13,7 @@ class User(BaseModel):
     email = db.Column(db.String(60), nullable=False, unique=True)
     image_url = db.Column(db.String)
     center_id = db.Column(db.String, db.ForeignKey('centers.id'))
+    role = db.Column(db.String)
```
GbengaOshinaga: @ibutiti There is a story to create roles for the application and it requires creating a different model for roles and associating it with the user model, so you should wait for the task to be completed and merged before finalizing your implementation.

ibutiti: Hey @GbengaOshinaga. Noted.

api/models/user.py
```diff
@@ -22,4 +23,4 @@ def get_child_relationships(self):
         return None
 
     def __repr__(self):
-        return f'<User {self.name}>'
+        return f'<User {self.name}, Role {self.role}>'
```
Dheavyman: Nice

api/views/center.py
```diff
@@ -44,3 +46,22 @@ def post(self):
             'message': SUCCESS_MESSAGES['created'].format('Center'),
             'data': center_schema.dump(center).data
         }, 201
+
+
+@api.route('/centers/<string:id>/people')
```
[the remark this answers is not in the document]
ibutiti: Had thought of adding this as a chore at some point, didn't reconcile the change you made with this :), thanks for pointing that out.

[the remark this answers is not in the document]
Dheavyman: You are welcome, but also know that this is a heads up in case my task gets merged before yours. But if yours gets merged first then the chore that will be created will take care of it.

api/utilities/helpers/random_string_gen.py
```diff
@@ -0,0 +1,12 @@
+"""Module for random string generator function."""
+import random
+from string import ascii_letters
+
+
+def gen_string(length=6):
```
frankip: I think you should also add where this function is used in the `Docstring` for future reference

seeders/seed_center.py
```diff
@@ -0,0 +1,16 @@
...
+    """Function to seed data in the center table.
+
+    :param total_records: The number of records to add.
+    :type total_records: int
+    """
+    for i in range(total_records):
+        db.session.add(Center(
+            name='Center '+gen_string(),
+            image={'url': 'www.cloudinary.com/'+gen_string()}))
+    db.session.commit()
```
Victoradukwu: Awesome. I choose how many seed centers I want to create.

Victoradukwu:
@ibutiti My bad. Ignore the above comment about returned users and message.

api/utilities/helpers/random_string_gen.py
Dheavyman: @frankip This method is for generating a random string which can be used anywhere as long as it matches your use case. Adding where it's been used here will not support its generic nature. That's my opinion, what do you think?

seeders/seed_center.py
Dheavyman: @Victoradukwu Thanks for the feedback. All changes necessary have been implemented. Please review

api/utilities/helpers/random_string_gen.py
frankip: Yeah that makes sense 👍

seeders/seed_center.py
```diff
@@ -0,0 +1,22 @@
...
+from api.utilities.helpers.random_string_gen import gen_string
+
+
+def seed_center(total_records):
+    """Function to seed data in the center table.
+
+    :param total_records: The number of records to add.
+    :type total_records: int
+    """
+    for i in range(total_records):
```
frankip: instead of using an unused variable, you can use an underscore`_` instead

Dheavyman: I researched the use of `_` for such case and I got to know that a `_` can be used when you don't really care about the value of the index. It seems like a common practice in python and hence might not hamper readability. I will change to it and we can go with it as long as no one has contrarily opinion. Thanks.

Dheavyman: Done

PR #75, opened 2018-07-30
#159350101 An endpoint to get all roles

appcypher:
#### What does this PR do?
- Have an endpoint to get a list of all roles

- An authenticated user should be able to get all the roles available and get a success message if all went well

- Create a local copy with the following commands: `git fetch origin ft-get-roles-159350101`, then `git checkout ft-get-roles-159350101`
- Drop the database with `dropdb activo` as a new model has been added and some other edited. 
- Create a new database: `createdb activo`
- Create migration file for role model: `flask db migrate`
- Add the tables to the database: `flask db upgrade`
- Using postman, send a GET request to: `localhost:5000/api/v1/roles`
- Run the tests with `pytest`

- N/A

[#159350101](https://www.pivotaltracker.com/story/show/159350101)

#### Postman collection:
[Link](https://www.getpostman.com/collections/14ace18903f7bc31031c)

#### Screenshot
screen shot 2018-07-30 at 1 32 59 pm


api/views/role.py
```diff
@@ -0,0 +1,37 @@
...
+    Resource class for creating and getting roles
+    """
+
+    @token_required
+    def get(self):
+        """
+        Gets list of all roles
+        """
+
+        roles = Role.query.filter_by(deleted=False).all()
```
Dheavyman: There is a generic _query method that does this query already, please checkout out the model operations.

appcypher: Will check that out. 👍

appcypher: Fixed!

api/utilities/model_serializers/role.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Role model schema
+    """
+    title = fields.String(
+        required=True,
+        validate=(string_length_60_validator, name_validator),
+        error_messages={'required': serialization_errors['field_required']})
+
+    @post_load
+    def convert_name(self, data):
```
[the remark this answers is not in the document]
appcypher: Fixed!

seeders/seed_roles.py
```diff
@@ -0,0 +1,21 @@
...
+
+from api.models import Role
+
+def seed_roles():
+    """
+    Seeds sample roles
+    """
+
+    roles = [
+        Role(title='Operations Intern'),
```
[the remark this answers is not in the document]
appcypher: Fixed!

api/utilities/validators/delete_validator.py
```diff
@@ -19,6 +19,7 @@ def delete_validator(relationships):
     if not relationships:
         return True
 
+    is_deletable = True
```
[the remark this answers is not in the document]
appcypher: That was put there because is_deletable will be undefined if `relationships` is empty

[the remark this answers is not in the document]
ibutiti: I think if there are no relationships, it returns true on line 20 and exits the function.

api/views/role.py
```diff
@@ -0,0 +1,37 @@
...
+        roles_data = RoleSchema(
+            many=True,
+            exclude=[
```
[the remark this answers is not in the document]
appcypher: True. Fixed! 🎉

api/utilities/validators/delete_validator.py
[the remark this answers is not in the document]
appcypher: After taking a closer look at the code, I noticed `relationships` won't be an iterable if there is just one relationship, therefore the `for .. in` path is never taken and the interpreter returns an error about `is_deletable` not being defined.

In my case, `role <-> user`, `relationships` isn't an iterable because there is only one relationship. 

I would like to have a sync with you concerning this.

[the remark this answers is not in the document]
appcypher: Thanks for the sync. It is now fixed! 🎉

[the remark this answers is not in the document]
ibutiti: Cool

Victoradukwu:
Nice job, @appcypher. Some feedback:
Endeavour to follow the team's commit message convention as documented in the project wiki. 
In the session **How should this be manually tested**, it is also important you include the need to run upgrade since database change is involved.

ibutiti:
Fix @Victoradukwu's comment though....

appcypher:
@Victoradukwu's fixed! :tada:

tests/test_role_endpoints.py
```diff
@@ -0,0 +1,32 @@
...
+        Test for getting assets
+        """
+        for role in new_roles:
+            role.save()
+        response = client.get(f'{BASE_URL}/roles', headers=auth_header)
+        response_json = json.loads(response.data.decode(CHARSET))
+
+        assert response.status_code == 200
+        assert 'deleted' not in response_json
+        assert 'deleted_at' not in response_json
```
purpose50: Awesome @appcypher , but since you are testing for a successful request you do not need to assert that those fields are not included. We expect them not to be included since you excluded them in your implementation. Also, including those fields means I can also assert for a field that does not relate in any way with this model,

E.g: ``` assert 'dancing_for' not in response_json ```

assert for the response only and only check if it contains what you expect

appcypher: Fixed! 🎉

appcypher: Fixed! 🎉

api/models/role.py
```diff
@@ -0,0 +1,28 @@
...
+        """
+        Method to get all child relationships of this model
+
+        Returns:
+             children(tuple): children of this model
+        """
+        return (self.users,)
+
+    def __repr__(self):
+        return f'<Role {self.id} {self.title}>'
```
Dheavyman: I will suggest you take out the self.id because in the case of creating a new role instance the id is none. The id is generated before the instance is added to the database. So it will only have value when you make a query to retrieve data from the database and not when you create a new instance. Having just self.title is enough and it's unique too.

tests/test_role_endpoints.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+from os import getenv
+from flask import json
+from api.utilities.constants import CHARSET
+
+BASE_URL = getenv('API_BASE_URL_V1')
+
+class TestRoleEndpoints:
+    """
+    Tests for getting assets
```
Dheavyman: This should read tests for getting roles instead I believe.

tests/test_role_endpoints.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    def test_get_assets_endpoint(self, client, init_db, auth_header, new_roles):
```
Dheavyman: I believe you meant test_get_roles... here. Also, add success to the name or something to signify this is for a success case.

tests/test_role_endpoints.py
```diff
@@ -0,0 +1,25 @@
...
+        """
+        Test for getting assets
```
Dheavyman: same as above on changing assets to roles

tests/test_role_endpoints.py
appcypher: Fixed! :tada:

tests/test_role_endpoints.py
appcypher: Fixed! :tada:

tests/test_role_endpoints.py
appcypher: Fixed! :tada:

brayoh:
@appcypher The postman collection should be a link

appcypher:
@brayoh Link fixed!

brayoh:
@appcypher fix the merge conflicts

appcypher:
@brayoh Fixed! 🎉

brayoh:
@appcypher the postman collection is not following the format we use, I'd suggest you pair with @GbengaOshinaga he shows you how we do the formatting.

appcypher:
@brayoh Postman collection fixed! 🎉

PR #87, opened 2018-08-08
#159557028 An endpoint to create a new role

kenoseni:
#### What does this PR do?
Allows a user to create a role

As a developer, I should be able to create a role by making a `POST` request to `/roles`

- pull the branch: `git fetch origin ft-create-new-role-159557028:ft-create-new-role-159557028`
- checkout to the branch: `git checkout ft-create-new-role-159557028`
- setup using commands: `pipenv shell
- upgrade to new structure: `flask db upgrade`
- run the app: `flask run`
- on postman: set bearer token and make a `POST` request to `/roles` to get a `201` response upon success
- make another `POST` request to the `/roles` route using the same data, to get a `409` error response
- make another `POST` request to the `/roles` route using invalid data, to get a `400` error response
- remove bearer token and make another get request to get a `401` error response
- **Request body Format**
```javascript
POST /roles
{
	"title": "operations manager",
	"description": "reports to the operations director"
}
```
- **Response body format**
```
{
    "status": "success",
    "message": "Role successfully created",
    "data": {
        "title": "Operations Manager",
        "description": "reports to the operations director",
        "id": "-LJoFaaPhq8UBCtD5E3f",
        "updatedAt": null,
        "createdAt": "2018-08-13T17:08:12.315874+00:00"
    }
}
```

N/A

[#159557028](https://www.pivotaltracker.com/story/show/159557028)

### Postman Documentation

[Run in Postman](https://app.getpostman.com/run-collection/f741b859d8bd21460c64)


api/views/role.py
```diff
@@ -2,8 +2,12 @@
 
 from flask_restplus import Resource
 from flask import jsonify
+from flask import request
```
rogerokello: Great work @kenoseni Consider making the import for request on the same line you made the import for jsonify ie from flask import jsonify, request

kenoseni: Thanks, @rogerokello, it makes sense.

api/views/role.py
```diff
@@ -32,3 +36,30 @@ def get(self):
...
         })
+
+    @token_required
+    @validate_json_request
+    def post(self):
+        """
+        Creates a role
+
+        Returns:
+            (dict): Returns status and success message
```
appcypher: Nice work! I believe this function returns more than just status and success message

kenoseni: @appcypher, thanks for seeing that. I will make the necessary corrections.

api/views/role.py
```diff
@@ -32,3 +35,31 @@ def get(self):
...
+            data(dict): Returns the id, title and description of the role created
+        """
+
+        request_data = request.get_json()
+
+        role_schema = RoleSchema(only=["id", "title", "description"])
```
GbengaOshinaga: @kenoseni Instead of passing `only`, make use of the `EXCLUDED_FIELDS` constant instead

GbengaOshinaga:
@kenoseni Fix the requested change and it'll LGTM, though the error that occurs when trying to upgrade the database needs to be resolved

GbengaOshinaga:
@kenoseni Trying to create or update a role that has been deleted throws an `Internal Server Error` because the role had been soft-deleted, but the unique constraint on its name still applies, therefore I suggest you drop the unique constraint on the `title` field in the role model

api/views/role.py
kenoseni: @GbengaOshinaga thanks for reviewing my work. The reason why I did not use the `EXCLUDED_FIELDS` was it will still return fields like `created_at` and `updated_at` in the response object which are not necessary. Updating `EXCLUDED_FIELDS` to include `created_at` and `updated_at` will affect the response object of other requests.

kenoseni:
@GbengaOshinaga Ok, thanks

api/views/role.py
GbengaOshinaga: @kenoseni Yeah, we usually return those fields, which was why they were excluded from the `EXCLUDED_FIELDS` constant. I advise you use it as is

kenoseni: @GbengaOshinaga  Ok, I will use it as is

kenoseni: Done

tests/test_role_endpoints.py
```diff
@@ -23,3 +28,139 @@ def test_get_role_endpoint(self, client, init_db, auth_header, new_roles):
         assert response.status_code == 200
         assert response_json['status'] == 'success'
         assert isinstance(response_json['data'], list)
+
```
appcypher: You did not test for _missing_ required fields in request body. 
You should test for missing `description` and/or `title` field, for example.

appcypher: I've seen where you implemented that.
I'll dismiss this review.

api/views/role.py
```diff
@@ -32,3 +36,31 @@ def get(self):
...
+        Returns:
+            (dict): Returns status and success message
+            data(dict): Returns the id, title and description of the role created
+        """
+
+        request_data = request.get_json()
+
+        role_schema = RoleSchema(exclude=EXCLUDED_FIELDS)
+        role_data = role_schema.load_object_into_schema(request_data)
+        validate_duplicate(Role, title=request_data["title"].title())
```
rogerokello: @kenoseni Consider doing the validation validate_duplicate from RoleSchema's convert_title method. I think it would be great to do most validation from the schema. It's just the way I look at it.

Victoradukwu:
@kenoseni  What is the update on the option of removing `unique` constraint from role name?

tests/conftest.py
```diff
@@ -53,34 +53,27 @@ def client(app):
...
+            description='reports to the operations manager'),
+        Role(
+            title='Operations Manager',
+            description='reports to the operations director'),
     ]
     return roles
 
 
 @pytest.fixture(scope='module')
 def new_role(app):
```
frankip: You should look for a more descriptive name for this fixture it looks similar to the _new roles_  fixture

frankip: It's okay since you are working on the function, you can just change the name and replace it where it's been used

kenoseni: @frankip Changing the name of this fixture will require making changes in a lot of files which are outside the scope of this task. If this fixture needs to be renamed then I suggest a new chore should be created to handle it so as to keep the file changes here within the scope of the task. This will also make the review faster and easier.

kenoseni: @frankip Taking a closer look to your earlier comment, I want to clarify that I did not create a new fixture but only added `description` to the existing ones.

api/models/role.py
```diff
@@ -11,9 +11,13 @@ class Role(AuditableBaseModel):
 
     __tablename__ = 'roles'
 
-    title = db.Column(db.String(60), nullable=False, unique=True) # pylint: disable=E1101
-    users = db.relationship( # pylint: disable=E1101
-        'User', backref='role', cascade='delete', lazy='dynamic')
+    title = db.Column(db.String(60), nullable=False)  # pylint: disable=E1101
```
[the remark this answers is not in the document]
kenoseni: The roles title cannot have a  unique constraint because of soft delete. This causes the server to crash when an attempt is made to add a role that has been soft deleted initially.

nzediegwu1:
LGTM

SEUNAGBEYE:
@kenoseni I suggest you should add the structure of the data to be sent in the `how should this be manually tested` section.

rogerokello:
Your response format looks like this
```
{
    "status": "success",
    "message": "Role successfully created",
    "data": {
        "title": "Operations Manager",
        "description": "reports to the operations director",
        "id": "-LJoFaaPhq8UBCtD5E3f",
        "updatedAt": null,
        "createdAt": "2018-08-13T17:08:12.315874+00:00"
    }
}
```

However the PT board respones format looks like this:
```
{ 
 data: [{ 
    id: '<role_id>,
    title: 'xxx',
    description: 'description of role'
  }],
  "message": "role created successfully",
  "status": "success"
}
```

My opinion is that it would not be necessary to supply other fields such as "updatedAt" or "createdAt" in the response, especially if it was not specified. I stand to be corrected.

kenoseni:
@rogerokello The response format you see in the PR is a modification of what is on the PT board. It is the convention that the team have been using, hence, the creation of EXCLUDED_FIELDS. Though it was not specified it makes sense to return it when thinking of the API as stand alone.

PR #88, opened 2018-08-08
#159533673 Refactor Load Method To Allow Partial Loading

GbengaOshinaga:


api/utilities/model_serializers/center.py
```diff
@@ -30,7 +32,9 @@ class CenterSchema(AuditableBaseSchema):
     @post_load
     def convert_name(self, data):
         """Convert center name with the first letter capitalized"""
-        data['name'] = data['name'].title()
+        if 'name' in data:
+            data['name'] = data['name'].title()
+            validate_duplicate(Center, name=data['name'])
```
Dheavyman: I suggest you take out line 36 and just convert the name to title-case here. This will make the code more concise.

GbengaOshinaga: @Dheavyman The line is there to convert the center name to title case before saving

PR #91, opened 2018-08-14
#158731109 Correction of the Endpoint to Update an Asset

ibutiti:
#### What does this PR do?
Corrects the endpoint to PATCH update an asset that was previously implemented as a PUT requiring the entire payload to be supplied.

The endpoint now can accept only the specific standard or custom attributes that are to be edited.

As a developer, I should be able to update an asset

Acceptance Criteria:
GIVEN I have valid authentication token
WHEN I send a patch request to the url /assets/{asset-id} (with any of the attributes in payload shape below with the exception of custom attributes which have to be nested inside the `customAttributes`),
THEN I will receive confirmation of the successful update of the asset

```
REQUEST
PATCH /assets/{asset-id}
{
   "tag": "gfredtjj76",
   "centerId": "IIUYHGRFfgdGH",
   "assetCategoryId": "-77YHTFRE",
   "customAttributes": 
      {
         "color": "Indigo",
         "warranty: "One year"
      }
}

RESPONSE
200 Ok
```
Create a local copy with the following commands: `git fetch origin ft-update-asset-correction-158731109:ft-update-asset-correction-158731109`, then `git checkout ft-update-asset-correction-158731109`
Start the application with the command:` flask run`
Using postman, send a PATCH request to: `localhost:5000/api/v1/assets/<asset_id>` with a payload in the shape as above.

1. Sending a value of `null` under any key will be interpreted as wanting to delete that attribute.
2. The centerId must be already registered in the db. This can be set to null.
3. The `assetCategoryId` must also already be registered in the db. This is however not nullable.
4. If a `customAttribute` has choices, the value passed in will be validated against these choices.
5. If a `customAttribute` is required, setting a null error will give a 400 response.

[#158731109](https://www.pivotaltracker.com/story/show/158731109)

N/A
#### Postman Collection
[Link](https://www.getpostman.com/collections/826ee60fb402b8a7fd5d)


api/utilities/model_serializers/asset.py
```diff
@@ -24,3 +27,28 @@ class AssetSchema(AuditableBaseSchema):
...
         validate=string_length_60_validator,
         error_messages={'required': serialization_errors['field_required']})
+
+    center_id = fields.String(
+        load_from="centerId",
+        dump_to="centerId",
+        validate=[string_length_60_validator, center_exists])
+
+
+class UpdateAssetSchema(AuditableBaseSchema):
```
GbengaOshinaga: @ibutiti This update schema is not necessary, use the `AssetSchema` and pass in `partial=True` to the `load_object_into_schema` method

api/views/asset.py
```diff
@@ -113,16 +113,83 @@ class SingleAssetResource(Resource):
...
-        asset.update_(  #pylint: disable=W0212
+        data = request.get_json()
+
+        # catch null value of center and remove center from asset otherwise
+        # serializer schema will reject it
+        if 'centerId' in data and data['centerId'] is None:
+            del data['centerId']
+            asset.center_id = None
+
+        schema = UpdateAssetSchema(partial=True)
```
GbengaOshinaga: `partial=True` should be passed into the `load_object_into_schema` method

api/utilities/model_serializers/asset.py
ibutiti: Hey @GbengaOshinaga, it is necessary because i'm using a different validator for the tag. When updating, we check whether there is another tag under an asset with a different ID

api/views/asset.py
ibutiti: @GbengaOshinaga It does the same thing but yeah, will change it for consistency.

ibutiti:
Hey @Victoradukwu, thanks for pointing that out. I needed to rebase. Now done and documentation updated accordingly.

appcypher:
You may need to amend your commit to have `feat` in its title instead of `feature`.

ibutiti:
Hey @appcypher. Thanks for that. Fixed. We may however need to harmonise our project wiki and the engineering handbook.

api/views/asset.py
```diff
@@ -113,16 +113,84 @@ class SingleAssetResource(Resource):
...
         asset = Asset.get_or_404(asset_id)
-
-        # asset_data_validators function parses and validates request data
-        asset.update_(  #pylint: disable=W0212
+        data = request.get_json()
+
+        # catch null value of center and remove center from asset otherwise
+        # serializer schema will reject it
+        if 'centerId' in data and data['centerId'] is None:
+            del data['centerId']
```
frankip: why should you delete if none is provided and since it's a patch my best recommendations is to leave it if null or none is supplied

tests/test_asset_patch_endpoint/test_asset_custom_attrs_patch.py
```diff
@@ -0,0 +1,149 @@
...
+from flask import json
+
+from api.models import Asset, AssetCategory, Attribute, Center
+from api.models.database import db
+from api.utilities.helpers.random_string_gen import gen_string
+from api.utilities.messages.success_messages import SUCCESS_MESSAGES
+from api.utilities.messages.error_messages import serialization_errors, jwt_errors
+
+
+@pytest.fixture(scope='function')
```
frankip: why add the fixtures to the actual test file, aint it supposed to  be done on the conftest?

tests/test_asset_patch_endpoint/test_asset_std_attrs_patch.py
```diff
@@ -0,0 +1,186 @@
...
+
+from api.models import Asset, AssetCategory, Center
+from api.utilities.helpers.random_string_gen import gen_string
+from api.utilities.messages.success_messages import SUCCESS_MESSAGES
+from api.utilities.messages.error_messages import jwt_errors
+
+BASE_URL = getenv('API_BASE_URL_V1')
+
+
+@pytest.fixture(scope='module')
```
frankip: move the fixtures to conftest file

tests/test_asset_patch_endpoint/test_asset_custom_attrs_patch.py
ibutiti: Fixed

tests/test_asset_patch_endpoint/test_asset_std_attrs_patch.py
ibutiti: Fixed

api/views/asset.py
ibutiti: Line 124 removes the key `centerId` from the request data if it is null. This is because marshmallow will raise an error when this happens, which we don't want because the `center_id` field is nullable.

The implementation does not delete/remove an attribute when a key is not provided in the request. For instance, if the db has the `center_id` set to something and the request does not have the `centerId` key in it, then the `center_id` is not altered.

As you rightly put it, when a key isn't there in a PATCH request, nothing is done to the resource in the db. This poses a challenge for us... How do we delete a nullable standard attribute or an optional custom attribute?

I discussed this at length with @SEUNAGBEYE and @brayoh and the best course of action was to send through the attribute we want to delete with the value null. If we were to delete the `center_id` we'd send the request through as:
```
{
"centerId" : null
}```

**Comment by appcypher** [comment]
LGTM

File: api/views/asset.py [hunk 6]
```diff
@@ -113,16 +113,84 @@ class SingleAssetResource(Resource):
     @token_required
     @validate_id
     @validate_json_request
-    def patch(self, asset_id):  #pylint: disable=C0103,W0622
-        """Endpoint for editing an asset."""
-
+    def patch(self, asset_id):
```
SEUNAGBEYE: @ibutiti Please refactor this function to reduce its cognitive complexity. The lines of code in this function is too long. Consider moving to validation to somewhere else.

ibutiti: @SEUNAGBEYE Fixed.

tests/test_asset_patch_endpoint/test_asset_custom_attrs_patch.py
```diff
@@ -0,0 +1,92 @@
...
+from api.utilities.messages.success_messages import SUCCESS_MESSAGES
+from api.utilities.messages.error_messages import serialization_errors
+
+BASE_URL = getenv('API_BASE_URL_V1')
+
+
+class TestAssetPatchEndpointCustomAttributes:
+    """Test class for Asset resource custom attributes in PATCH endpoint."""
+
+    def test_patch_asset_null_required_custom_attr(
```
SEUNAGBEYE: @ibutiti I should be able to look at a test method and have a clue of what it does. Why not try something like this: `the name of the tested feature]_[expected input / tested state]_[expected behavior]` (`test_update_asset_setting_required_atrr_to_null_should_fail`)

tests/conftest.py
```diff
@@ -366,3 +367,71 @@ def test_center_with_deleted_users(new_role):
...
+            'screen size': 'Red',
+            'color': 'Green'
+        }
+    }
+    asset = Asset(**data)
+    return asset.save()
+
+
+@pytest.fixture(scope='module')
+def asset_no_attrs():
```
SEUNAGBEYE: @ibutiti I think `asset_without_attrs` is more descriptive than `asset_no_attrs`. What do you think?

ibutiti: @SEUNAGBEYE Agreed. On it.

tests/test_asset_patch_endpoint/test_asset_custom_attrs_patch.py
ibutiti: On it @SEUNAGBEYE

ibutiti:
@SEUNAGBEYE, are there any other changes you'd like done on the PR so I can execute them altogether?

SEUNAGBEYE:
@ibutiti No.

PR #92, opened 2018-08-15
#159557240 Get a list of all permissions

appcypher:
#### What does this PR do?
- Have an endpoint to get a list of all permissions

- An authenticated user should be able to get all the permissions available and get a success message if all went well

- Create a local copy with the following commands: `git fetch origin ft-get-all-permissions-159557240`, then `ft-get-all-permissions-159557240`
- You need to add seed data to database test the api: `flask seed_database`
- If that doesn't work, then you need to recreate the database:
    * Drop database: `dropdb activo`
    * Create database: `createdb activo`
    * Add the tables to the database: `flask db upgrade`
    * Populate database with seed data: `flask seed_database`
- Using postman, send a GET request to: `localhost:5000/api/centers/permissions`
- Run the tests with `pytest`
- **Response body format**
```javascript
{
	"data": [{
		"id": "-LJxPmlLDbK41JrqmMR0",
		"type": "Delete"
	}],
	"message": "Permissions fetched successfully",
	"status": "success",
	"meta": {
		"currentPage": "http://localhost:5000/api/v1/permissions",
		"firstPage": "http://localhost:5000/api/v1/permissions?page=1&limit=10",
		"nextPage": "",
		"page": 1,
		"pagesCount": 1,
		"previousPage": "",
		"totalCount": 5
	}
}
```

- N/A

[#159557240](https://www.pivotaltracker.com/story/show/159557240)

#### Postman collection:
[Link](https://www.getpostman.com/collections/3797c53f51a435efa52b)

#### Screenshot
- N/A


rogerokello:
1. Great work @appcypher, In your PR description under the section `How should this be manually tested?`, the first bullet, did you mean `git checkout ft-get-all-permissions-159557240` or just ft-get-all-permissions-159557240?

2. I've also looked at the response body format as specified in the PR description section `How should this be manually tested?` and compared it with what is in your postman collection examples. There is a  meta key in the response that does not seem to appear in the postman collection examples. Looking into that may be necessary too.

appcypher:
@rogerokello 
**1.** Fixed! 🎉 

**2.** According to @SEUNAGBEYE, the response format on the PT board is not `hard-set` and the final response implementation may contain additional data as appropriate. 

That is not to say the response format should not be adhered to, however, there are special cases where the extra information is needed. In this case, the pagination helper I added to `get all roles` implementation generates an object that is considered useful and meant to be added to the response body.

I hope you understand my point.

tests/test_permision_endpoint.py
```diff
@@ -0,0 +1,77 @@
...
+            new_permissions(BaseModel): fixture for creating a permissions
+        """
+        for permission in new_permissions:
+            permission.save()
+        response = client.get(f'{BASE_URL}/permissions', headers=auth_header)
+        response_json = json.loads(response.data.decode(CHARSET))
+
+        assert response.status_code == 200
+        assert response_json['status'] == 'success'
+        assert response_json['message'] == \
```
ibutiti: Validate the count of the permissions with the length of the fetched list.

appcypher:
@ibutiti Description fixed! 🎉

tests/test_permision_endpoint.py
appcypher: Fixed! 🎉

api/utilities/paginator.py
```diff
@@ -43,23 +43,23 @@ def validate_pagination_args(arg_value, arg_name):
         })
 
 
-def pagination_helper(model, schema, extra_query=None):
+def pagination_helper(model, schema, extra_query=None, exclude=EXCLUDED_FIELDS, only=None):
```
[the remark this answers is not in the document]
appcypher: `pagination helper function` is extended to allow excusion/inclusion of certain fields from its output data.
`only` is made None by default so that it won't be applied to the returned data, if `only` were to be an empty array by default, then nothing will be returned.

appcypher:
@ibutiti 
Tests written for pagination

api/utilities/paginator.py
nzediegwu1: this line seem to violate the max line length. consider refactoring like: `def pagination_helper(model,
                      schema,
                      extra_query=None,
                      exclude=EXCLUDED_FIELDS,
                      only=None):
`

nzediegwu1:
In your `How should this be manually tested`, is the instruction to generate migration file necessary? Consider removing since migration file has already been created

appcypher:
@nzediegwu1 
It is needed, permission migration file needs to be generated

api/utilities/paginator.py
appcypher: Fixed! 🎉

api/views/permission.py
```diff
@@ -0,0 +1,36 @@
...
+        Gets list of all permissions
+        """
+
+        permission_data, pagination_object = pagination_helper(
+            Permission,
+            PermissionSchema,
+            only=['id', 'type']
+        )
+
+        return jsonify({
```
nzediegwu1: Nice work. Use of jsonify here seem redundant since flask-restplus being used has taken care of it. Consider returning just the dict

tests/conftest.py
```diff
@@ -7,7 +7,7 @@
 from main import create_app
 from config import config
 from api.models.database import db as _db
-from api.models import User, Asset, AssetCategory, Attribute, Center, Role
+from api.models import User, Asset, AssetCategory, Attribute, Center, Role, Permission
```
nzediegwu1: This line seem longer than the max-length. Consider breaking it

api/views/permission.py
appcypher: Fixed 🎉

tests/conftest.py
appcypher: Fixed 🎉

api/views/permission.py
```diff
@@ -0,0 +1,36 @@
...
+    Resource class for creating and getting permissions
+    """
+
+    @token_required
+    def get(self):
+        """
+        Gets list of all permissions
+        """
+
+        permission_data, pagination_object = pagination_helper(
```
nzediegwu1: Is pagination needed here? considering that this resource has a length of about four items. @SEUNAGBEYE, @andela-kakpobome, please look into this

appcypher: Will remove

api/views/permission.py
```diff
@@ -0,0 +1,36 @@
+"""Module for permission resources"""
+
+from flask_restplus import Resource
+from flask import jsonify
```
rogerokello: Great work @appcypher, it may be necessary to create a space here to separate Library imports from module imports.

rogerokello: Thanks, @appcypher, it is not a convention however it would help for easier readability.

api/views/permission.py
appcypher: Fixed 🎉

api/views/permission.py
appcypher: Fixed 🎉

api/models/permission.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Permission(AuditableBaseModel):
+    """
+    Model for permissions
+    """
+
+    __tablename__ = 'permissions'
+
+    type = db.Column(db.String(60), nullable=False, unique=True) # pylint: disable=E1101
```
kenoseni: @appcypher Good job for implementing this task. You created a variable name on line 14 called `type`. `type` is a keyword in `Python` so I suggest you consider changing it.

api/utilities/model_serializers/permission.py
```diff
@@ -0,0 +1,18 @@
...
+from ..validators.name_validator import name_validator
+from ..messages.error_messages import serialization_errors
+from ..validators.string_length_validators import string_length_60_validator
+
+
+class PermissionSchema(AuditableBaseSchema):
+    """
+    Permission model schema
+    """
+    type = fields.String(
```
kenoseni: the comment above also applies to this line

api/models/permission.py
appcypher: I think I will stick with `type`. Since there is no potential clash with Python's builtin `type` class.
`type` in this case, as well as `id`, is being used as a field not a class.

It will only be referenced using dot notation, as in `Permission.type`.

The same issue applies to other classes, we can't say because we have class `greet` defined, we can't have a field named `greet`

appcypher: @SEUNAGBEYE ⬆️

PR #93, opened 2018-08-16
#159749708 Refactor validate duplicate function

kenoseni:
#### What does this PR do?
Refactor the existing validate duplicate function

When updating any resource with the same data, a conflict error must not occur as a response

- pull the branch: `git fetch origin bg-refactor-validate-duplicate-159749708:bg-refactor-validate-duplicate-159749708`
- checkout to the branch: `bg-refactor-validate-duplicate-159749708`
- setup using commands: `pipenv shell
- upgrade to new structure: `flask db upgrade`
- run the app: `flask run`
- on postman: set bearer token and make a `PATCH` request to `/center/<center_id>` without changing the request body for updating a center to get a `200` response upon success
- make another `PATCH` request to the `/asset-categories/<asset-category_id>`, without changing the request body for updating an asset category to get a `200`  response upon  success

When updating a resource with the same data, the response object returns an error message that the resource already exists in the database instead of returning the data. This is caused by the initial implementation of the `validate duplicate ` function which is now refactored to fix this bug.


tests/test_update_centre_endpoint.py
```diff
@@ -8,7 +8,7 @@
 
 from api.utilities.constants import CHARSET
 from api.utilities.messages.error_messages import serialization_errors, jwt_errors
-from .mocks.center import VALID_CENTER
+from .mocks.center import VALID_CENTER, NEW_CENTER, INVALID_CENTER, VALID_CENTER_TWO, INVALID_CENTER_NAME
```
nzediegwu1: This line is longer than the specified length in our convention. consider refactoring

api/utilities/validators/duplicate_validator.py
```diff
@@ -14,7 +14,18 @@ def validate_duplicate(model, **kwargs):
...
     """
 
-    result = model.query.filter_by(deleted=False, **kwargs).first()
+    record_id = kwargs.get('id')
+
+    kwargs.pop('id', None)
+
+    query = dict(deleted=False, **kwargs)
+    if record_id:
+        result = model.query.filter_by(**query).filter(
```
rogerokello: Great work @kenoseni, thanks for the refactoring of the tests I wrote. Just one comment; My thoughts are that for future reference, it would be necessary to probably give an inline comment specifying what you are trying to achieve here. Thank you.

api/utilities/validators/duplicate_validator.py
```diff
@@ -14,7 +14,18 @@ def validate_duplicate(model, **kwargs):
          kwargs(dict): keyword arguments containing fields to filter query by
     """
 
-    result = model.query.filter_by(deleted=False, **kwargs).first()
+    record_id = kwargs.get('id')
+
+    kwargs.pop('id', None)
+
+    query = dict(deleted=False, **kwargs)
+    if record_id:
```
rogerokello: Great work @kenoseni, thanks for the refactoring of the tests I wrote. Just one comment; My thoughts are that for future reference, it would be necessary to probably give an inline comment specifying what you are trying to achieve here. Thank you.

SEUNAGBEYE: @kenoseni +1 on this

tests/test_update_centre_endpoint.py
kenoseni: Thanks, @nzediegwu1, I will make the necessary corrections.

api/utilities/validators/duplicate_validator.py
kenoseni: Thanks, @rogerokello, I will update the code with inline comments.

